PR #8, opened 2018-08-21
added kagome lattice

mrader1248:


jhauschild:
First of all: Thanks a lot!

Before I merge it, a possibly stupid question from my side:
Does the parameter `x` really change something?
Clearly, you've rotated the lattice such that for `x=True`, a bond is aligned along the x-axis of the "plot", i.e. in the coordinate system in which the `basis` is given; for `x=False` we have a bond along the y-axis. But does that really change anything physically, or is it just a rotation in the "plot" coordinate system?
The `CouplingModel` without a shift connects the sites shifted by `Ly*basis[1]` with the site at `(0,0)`, so these seem to be the same sites, independent of `x`. As far as I see it, we get what's called `YC2n-0` in  [PRX 7, 031020 (2017)](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.7.031020) when using `bc_coupling='periodic'` for both `x`.  Or am I missing something? 
Of course the ordering of the sites is different, but that's what the 'ordering' parameter should be used for... Also, I think it's a bit confusing that changing the `x` parameter changes the number of sites in the lattice unit cell.
In summary, do we really need the `x` parameter?

mrader1248:
If you are only dealing with finite-size simulations, then it doesn't matter whether you set `x` to `True` or `False`. However, if you are dealing with infinitely long cylinders, it will change the boundary conditions and depending on the physical model it will also change the results (for finite `Ly`).

jhauschild:
Well, what matters is which sites you identify when going around the cylinder, right?
By shifting the sites identified, i.e., the `m` in the YC2n-2m of the above PRX, we can "rotate" the lines of allowed momenta in the Brillouin zone, compare Fig. 1 c) and d).
Rotating enough, as mentioned in the paper, we get the same lines with YC2n-n as with XC2n in Ref 8 of the paper, or maybe more generally XC2n-0.

Now we need to be careful which sites are identified by our tenpy code (which happens in the `CouplingModel`). As I wrote, for a shift `m=0` the code identifies the sites located at `Ly*basis[1]` with the one at `(0, 0)`. I've marked identified sites with the big orange marker in the following image:

kagome

The cylinder axis goes perpendicular to the line connecting the two orange dots - and that's the same in both lattices, if you rotate the whole "plot" on the left by 30 degrees. What is different is only how you cut the cylinder open to get it into the plane, but that shouldn't make a big difference, does it?
(I'm claiming that the you also get a YC2n-0 lattice for `x=True`, as we also see in Fig. 1 a) of the PRX.)
That's why I think they are equivalent in the end?

mrader1248:
Okay, I see your point. Confused myself quite a bit with these BCs. Does it matter for the computational complexity (of DMRG, TEBD, ...)? If it doesn't matter, I will adapt my version and drop the 6-site unit cell. Also, I think examples for constructing XCn-m and YCn-m should be included.
However, what is still not obvious to me: The BCs are part of the geometry (and therefore of the lattice), not of the model, whereas in this library, they are coming from the model?

mrader1248:
I removed the 6-site unit cell, so I guess it's ready to merge now.

Regarding the boundary conditions: I don't think it makes sense to talk about "coupling BCs". For instance, one might have a model coupling NNs, but the question "Which sites are NNs of site xxx?" should already be answered by the geometry. On the level of MPS/MPO of course everything is purely one-dimensional.

jhauschild:
Ok, thanks a lot!
I called it `bc_coupling` to distinguish it clearly from the `bc_MPS` and to make people aware that these are different bc; "coupling"  because it's used in the `CouplingModel`.
We could move the whole `bc_coupling` into the lattice and call it `bc_lattice`.

mrader1248:
Thank you for merging my code!

I would strongly advise you to move the boundary conditions completely from the model to the lattice, as this separates model and lattice logic more clearly.

PR #10, opened 2018-09-30
lattice-independent model definitions

mrader1248:
I added two (very simple) classes `TFIModel` and `XXZModel`, which can be used to define these models for any lattice (defining the required property `nearest_neighbors`).

For example, the TFI model can be defined using
```
model = TFIModel(lattice, 0.5)
```
where `lattice` can be defined by each of the following statements.
```
lattice = Chain(L, site)
lattice = Square(Lx, Ly, site)
lattice = Honeycomb(Lx, Ly, site)
lattice = Kagome(Lx, Ly, site)
```


jhauschild:
Thanks for this contribution!
Did you realize that the `spins.TFIModel` you defined is fully equivalent to the `tenpy.modes.tf_ising.TFIModel2D`?
You can call the `TFIModel2D` like this, if you like:

    lat = Square(Lx, Ly, site) # or whichever lattice you like
    model_pars = {'lattice': lat, 'g': 0.5, 'J': 1}
    model = TFIModel2D(model_pars)

I don't think it makes sense to keep two versions of the exact same model side by side. Or do you feel that the `TFIModel2D` is already too complicated as an example? It is included into the userguide "Introduction to models", so if it is unclear, we should add a few more explaining sentences there.

For the XXZModel, we should probably directly generalize the SpinChain to allow arbitrary lattices.

We can also continue our [discussion in the forum](https://tenpy.johannes-hauschild.de/viewtopic.php?f=9&t=7).

mrader1248:
No, I only knew the old version, which defined the lattice internally. But the big question is: Is `TFIModel2D` really a 2D model? I could also create it on top of a `Chain` and it turns into a 1D model. Or is there any advantage in inheriting from the `NearestNeighborModel`?

To be honest: For me TFI just means to have couplings on NN-bonds and a transverse field on each site -- independent of the lattice. I could imagine it's much easier for beginners to just declare the model on top of a lattice. Otherwise it suggests that each model has to be implemented independently for each lattice. In addition it reduces the amount of code in tenpy without reducing the number of features.

If you only keep one class for the XXZ model, I would as well only keep one class for the TFI model (to be consistent), and rather shift the `TFIChain` class to the example codes.

jhauschild:
Yes, there is a need for deriving from the `NearestNeighborModel`: it states that you can write the hamiltonian as sum of two-site bond terms, nearest neigbor in the sense of the DMRG/MPS snake (= lattice 'order'). This information is crucial for the Suzuki-Trotter decomposition in the TEBD algorithm, TEBD only works on NearestNeighborModels. So at least for the moment (where we don't have an time evolution working directly with an MPO implemented yet), this is kind of model for which you can do time evolution...
Strictly speaking, what TEBD checks/needs is only that there is an `H_bond` attribute, and not whether the class is derived from the NearestNeigborModel. So a hack would be to just set the H_bond attribute if the lattice is a `Chain`. Yet, sooner or later, this will lead to confusion what models can be used for TEBD.
Instead, I suggest [this adjustment](https://tenpy.johannes-hauschild.de/viewtopic.php?f=9&t=7&p=28#p28).

mrader1248:
Okay, now I see, why there is this `NearestNeighborModel`. But in the end at some point somebody might implement TEBD for MPOs or even TDVP and then any MPO could be used. Therefore maybe it is more user-friendly if the models were merged (with a possible `H_bond` attribute) and if one tries to use TEBD with a non-NN model, an error should be raised (but of course it should also be in the documentation of TEBD that a NN model is needed).

And sorry, I ignored your discussion in the forum. At some point I thought my post vanished and I didn't track it any longer.

jhauschild:
I'm still in favor for different model classes for the different ways to represent the Hamiltonian, for the following reason:
Having the different classes allows to implement methods for the conversion between the different representations in a very structured way. Right now there is only the conversion from the CouplingModel to the MPO or NN model. In general, it should also be fairly easy to convert the NNmodel bond-representation into a MPO with a bunch of SVDs, such a function can now be implemented in the NNModel. 
Imagine that someday someone will write a PEPS code, which in a simple version (e.g. for the simplified update) requires a representation of H in terms of nearest-neigbor bonds in the 2D lattice (which become long-range bonds in 1D). He can then just create a new model class for yet another representation, and if possible implement some conversion functions from the other representations.
Having these conversion functions is the real trick by which you avoid rewriting the same model again and again for different algorithms...

> And sorry, I ignored your discussion in the forum. At some point I thought my post vanished and I didn't track it any longer.

I moved the discussion into a different subforum/category; I'm sorry if it looked like I just deleted your post. I'm still learning about the forum software and should have kept a "shadow" version in the original forum.

mrader1248:
I see that there is no total order of the inheritance relation between `MPOModel` and `NearestNeighborModel` -- especially when it comes to 2D. However, this whole idea of mixins requires to implement the same model for each combination of dimensionality and ansatz, e.g. something very simple like the TFI model:

* 1D + MPS ==> `TFIChain(MPOModel, NearestNeighborModel)`
* 2D + MPS ==> `TFI2D(MPOModel)`
* 2D + PEPS ==> `TFIPEPS(NearestNeighborModel, PEPOModel)`

If there would be just one base class for any model (handling the management of the couplings) offering methods like `get_mpo`, `get_pepo`, ... (which could of course raise some error if this operation is not supported/not possible), you would circumvent this jungle of classes for one single model.

jhauschild:
Thanks again for the suggestion!
Although the changes did not get merged directly, I've followed your idea and now defined most of the models in tenpy based on the new class `CouplingMPOModel`, which by default accepts any lattice for the `lattice` model parameter.
I allow myself to close this pull request without merging it (sorry about that!), and refer once more to the [forum](https://tenpy.johannes-hauschild.de/viewtopic.php?f=9&t=7&p=61#p61) for further discussions

PR #16, opened 2019-01-02
additional utility functions in mps.py / mpo.py

gunnarmoller:
-utility functions to create new bunched mps/mpo objects
-utility function to calculate average charge


jhauschild:
Hi Gunnar, thanks for your contributions!

Before we merge these functions, I have some comments and questions.

Did you realize that the functions `MPS.get_bunched_mps()` and `MPO.get_bunched_mpo()` are already included in TeNPy? The `MPS.group_sites()` and `MPO.group_sites()` do exactly the same thing (apart from using the `GroupedSite` to combine the sites). I think we should avoid having them two times implemented.

The `average_charge` method looks very good, thank you very much!

For the `get_full_hamiltonian` and `full_contraction`, I'm a bit confused why you take the trace for the open indices on the left and right. For a finite system with only trivial MPO legs (only 'R' on the very left, only 'F' on the very right), it might work as long as the charges fit together, but in any other case, the trace isn't really what you want,  at least not if you think of Hamiltonian-like MPOs. (It might be different if you consider exp(Ht) or similar stuff).
I think what you actually want is to calculate something like the energy E = <psi|H|psi> by a full contraction, right?
This is better done with `energy = MPOEnvironment(psi, H, psi).full_contraction(0)`.
Note that for an infinite system, this will return the energy of all the terms *inside* the unit cell, *excluding* the terms between unit cells, and hence is different from the energy per site in the infinite system! The MPOEnvironment takes optional arguments firstLP, rightRP, which define the tensors to be used on the very left and right.

Finally, there are some minor notation changes between the previous (closed source) tenpy and this one: the labels of your new `full_contraction()` are not completely consistent with the rest of TeNPy. As far as I see, the 'b' and 'b*' should be vL and vR of MPS.get_B tensors, the 'w' and 'w*' of the MPO.get_W should be wL and wR, and the 'R' and 'F' indices from the finite state machine are now given by the IdL[0] and IdR[-1] attributes of the MPO.
Moreover, The `rename_label` should be a `replace_label`.
A good way to avoid such minor bugs due to convetion changes and typos would be to write a small test executing the newly implemented functions. It will greatly help us to always keep the latest development version working.

Finally, let me assure you that I don't wont to demotivate you from contributing with this lengthy criticism; I greatly appreciate your contribution! I only try to keep a high quality code base.
You can commit  further changes and include them into this merge request. Or do you want me to make the necessary changes before merging?

PR #38, opened 2019-06-11
'shelve' parameter should be set as an attribute,…

Lschoon:
… to allow other code to check if run ended because it ran out of time.

Note this is a _very_ minor change. However, I am working on some code to shelve and restart simulations without changing any of the parameters. Usecase for example: code run on a machine which limits job run time. I believe this is not implemented at the moment (please confirm). This could potentially be rolled into this pull request once done.


jhauschild:
Thanks!

PR #40, opened 2019-06-18
Updated and moved build_initial_state()

Lschoon:
The function is now general (can take any type as local state) and defined in `tenpy.networks.mps` rather than `tenpy.tools.misc`.


jhauschild:
Thanks a lot! I'd ask you for one more thing, though: please remove the old code from tenpy.tools.misc after the deprecation warning and just import and call the new function, as I did with chi_list.

PR #81, opened 2019-07-25
Chern Insulator Examples (3 new models + minor improvements)

bartandrews:
Hi Johannes,

I have some additions to the repository that others may find useful.

1) I have replaced external_flux.py from the advanced examples, with a "chern insulators" directory in the examples folder. In this folder, I include 3 (confirmed working) example chern insulators: the haldane model (https://arxiv.org/abs/1407.6985), the chiral pi flux model (https://arxiv.org/abs/1012.4723), and the generalized (C=3) Haldane model (https://arxiv.org/abs/1205.5792). For each of these we calculate the charge pumping and spectral flow, as before. The corresponding models are now in the models directory, for general use.

I have changed V=1 to V=0 in the Haldane model example, so that it exactly reproduces Fig.2.a,b) from the paper. 

You can run each of these examples simultaneously - they should take ~30 minutes to complete with the current parameters (haldane, chiral pi flux, generalized haldane - in order of speed). You can also uncomment a line in the files if you want to plot the hoppings. For the chiral pi flux model, we observe charge pumping with a Chern number of 1, which argees with literature. For C=3 Haldane model, we observe a charge pumping of 3 unit charges, which corresponds to a Chern number of 3, also in agreement with the literature. In all cases, we observe spectral flow - as expected. 

Note: For the cases, of chiral pi flux model and generalized haldane model, there are custom lattices included in the model file. One could argue that these custom lattices should be separate in the lattice directory. However, I think that they are not generic enough to be separated out. They merely serve as an example of how one can construct a custom lattice for a specific model.

2) I have added 4th and 5th nearest neighbors for the Honeycomb lattice because they are useful for people studying twisted bilayer graphene.

Please let me know if anything is unclear.

All the best

Bart


jhauschild:
Thank you very much for these examples, they are great! :+1: 

I requested a few minor fixes in a pull request to your branch. If you merge them, they will automatically appear hear.

Moreover, I feel like the model folder is getting a bit unstructured.
I would suggest to
- combine the files `fermions_haldane.py` and `fermions_haldane_C3.py` into a single file called `haldane.py` (I don't think we need to mention the fermion explicitly...)
- move the class definitions from `fermions_pi_flux.py`  directly into the `examples/chiral_pi_flux.py` file. They are very particular to this example, so I think it's fair to have them not as general models inside tenpy. (We could do the same for the `FermionicC3HaltandeModel` as well... But the general Haldane model for sure is usefull inside `tenpy.models`)

Finally, I'd like you to add a class doc-string to the model classes defined within `tenpy.models`.
These doc strings appear in the online documentation of the class (which appears on a page not containing the module doc-string). Hence the class doc string should at least reference the papers again such that the Hamiltonian becomes clear.

bartandrews:
Thank you for the fixes in the pull request, and for the useful corrections. I have made all of the necessary changes now :-)

As I was making corrections, there were a few things that I noticed:

1) There is such a thing as a bosonic haldane model, and so we may want to keep the name "fermions_haldane.py" to keep with the convention of "fermions_hubbard.py" and "bose_hubbard.py". On the other hand, "haldane.py" fits with the convention of "hofstadter.py", which has both fermionic and bosonic cases within one file. Is the naming convention inconsistent at the moment?

2) Why in the class doc string of FermionicHubbardModel does it say Spin-1/2 fermionic Hubbard model in 1D ? Specifically, why the 1D? You can also use this in 2D on a cylinder, right? Otherwise, you could argue that all DMRG is in 1D.

3) I think that the staggered chemical potential in the Haldane model was actually not a typo, so I have changed it back. It is:

self.add_onsite(mu, 0, 'N', category='mu N')
self.add_onsite(-mu, 1, 'N', category='mu N')

I think that the 1 in the second line means that it is acting on the B site.

examples/chern_insulators/haldane_C3.py
```diff
@@ -57,6 +57,10 @@ def init_sites(self, model_params):
         fs = FermionSite(conserve=conserve)
         gs = GroupedSite([fs, fs], labels=['A', 'B'], charges='same')
         gs.add_op('Ntot', gs.NA + gs.NB, False)
+
+        print(sorted(gs.opnames))
```
jhauschild: We should remove any print statements, or at least enclose them in a `if verbose > 1:` clause

jhauschild:
Thanks for implementing the requested changes! Regarding your comments:

1. That's a good point and indeed inconsistent, we should join the Fermi-Hubbard and Bose-Hubbard model into a single file `hubbard.py`. Since some people might be using it, we should keep it backwards compatible, though, and issue a warning about the movement. Similar as in the `bose_hubbard_chain.py`. That one should be adjusted as well...
2. That's a relict of the time when it was only defined in 1D. You're welcome to change it ;-)
3. The staggered potential is there in the haldane model right now, correct? We need it for the various programs.

As far as I see, the examples actually don't involve interactions, or do they? Should we mention this somewhere? The real power of DMRG is that everything still works if we add interactions.

One last thing: I think that the fourth- and fifth-nearest neighbors will not often be needed. So I'd suggest to instead just set them as `self.fourth_nearest_neighbors = ...` inside the `Honeycomb.__init__` instead of introducing the extra keywords in the `Lattice.__init__`. Do you agree?

bartandrews:
1. Done
2. Done
3. Yes, it's correct now

I have added an interaction term to all of the examples now. However, the default is still V=0, which is how they are intended to be run. As I mentioned before, unfortunately I cannot make it work with interactions...

Agreed, I have changed this.

examples/chern_insulators/haldane_C3.py
bartandrews: These print statement have been removed.

jhauschild:
Thanks a lot for implementing all this!

bartandrews:
Glad that I could contribute something :-) 

I just realized that I forgot to add "'JW', True" to the end of all my interaction couplings in "chiral_pi_flux.py", "haldane_C3.py", and "models/haldane.py". Does this matter? If so, could you add this on for me?

jhauschild:
The interactions are `N N` terms and don't need an Jordan-Wigner string, adding it would make it wrong.
(And by the way, `add_coupling` is adding it if necessary, depending on the operators, so no, you don't have to add it manually)

bartandrews:
Thanks for clarifying :thumbsup:

Issue #65, opened 2019-07-08
Literature list


jhauschild:
- [x] It's probably helpful to add the doi for all of the papers with the `:doi:` tag.
- [x] Maybe we should sort the literature by topics
- [x] add a few more references to general reviews and to papers introducing different methods.
- [x] The literature list in the HTML documentation is formated badly.
This seems to be a sphinx issue, can we circumvent this?

jhauschild:
ee4039880641b2c907de201d9bd38c9b9b548f19 fixed the formatting problem by accident, at least mostly. There is no space after the reference name, but I think that's fine.
It seems like Sphinx interpreted the abbbreviated first names as lists...

Issue #44, opened 2019-07-06
add missing __all__


jhauschild:
The` __all__ = [...]` at the beginning should contain the names of (public) classes and functions defined in a file. Check that each file in `tenpy/` contains this field with all required classes.
On linux, this command might help: 
`grep -L "__all__" $(find tenpy -name "*.py")`

PR #83, opened 2019-08-08
bosonic Haldane model & FCI benchmark (+ issues #44, #65)

bartandrews:
Hi Johannes,

I have made the following changes that may be helpful:

- fixed typo from my previous pull request (to make the plot couplings work for the haldane CI example)
- added bosonic Haldane model to haldane.py
- added bosonic Haldane FCI benchmark example. This coincidentally gives the correct FCI charge pumping with the current parameters (you can try it to see that one charge quantum is pumped after 4*pi flux insertion, which characterizes a 1/2 filled bosonic FCI), however it is not working at higher chi for example, or larger MPS unit cells... I think it is nevertheless useful as a benchmark because it is one of the simplest/cheapest FCI examples that should definitely work, and this example does show some signs of working. It could be useful in debugging the convergence issue with the DMRG engine.
- Literature list #65 (I am not sure if my rst comments are good practice - we can remove them if not)
- add missing __all__ #44 (checked recursively in the tenpy directory only)

All the best

Bart

### DIFF
--- a/doc/literature.rst
+++ b/doc/literature.rst
@@ -6,20 +6,70 @@ They can be cited from the python doc-strings using the format ``[Author####]_``
 If you're looking for introductory notes, don't forget the 'official' [TeNPyNotes]_. The review by [Schollwoeck2011]_ is
 also a classic introduction.
 
+**General reading**
 
+.. Density matrix formulation for quantum renormalization groups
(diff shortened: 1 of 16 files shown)


jhauschild:
Wow, that's great! Thanks a lot!

bartandrews:
No problem, I'm glad it came in useful :+1:

PR #91, opened 2019-10-01
Implementing MPO to MPS application and MPS compression

jhauschild:
@scaletos has been working on functions to apply an MPO to an MPS and compress it, as well as on generating the W_I/W_II approximations for exp(iHdt) for a given MPO H.

First of all, a big **thank you** for implementing this feature :+1: 

To simplify looking at the changes, I created this pull request.

Further things to be done:

- [x] does it work with non-trivial charges? Make it work....
- [x] add tests to check that everything works
- [x] variational MPS compression as faster alternative (probably to be postponed)
- [x] similar call structure as for TEBD/TDVP? reasonable?


tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    'make_U', 'make_UI', 'make_UII', 'make_WII', 'mps_compress', 'svd_two_site', 'apply_mpo'
+]
+
+def make_U(H, dt, which='II'):
+    r"""Creates the UI or UII propagator for a given Hamiltonian.
+
+    Parameters
+    ----------
+    H : :class:`~tenpy.networks.mpo.MPO`
+    dt : float|complex
```
jhauschild: copy over the (edited) comments from `make_UI`/`make_UII`.

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    mps.set_B((i+1)%mps.L, vh, form='B')
+    mps.set_SR(i, s)
+
+
+def apply_mpo(psi, mpo, trunc_par):
+    """Applies an mpo and truncates the resulting MPS using SVD.
+
+    Parameters
+    ----------
+    mps : :class:`~tenpy.networks.mps.MPS`
```
jhauschild: `mps` is called `psi` in the function arguments...

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+        MPS to apply operator on
+    mpo : :class:`~tenpy.networks.mpo.MPO`
```
jhauschild: let's rename the function argument  `mpo` -> `U_mpo`.

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+        MPO to apply. Usually one of make_UI() or make_UII(). The approximation being made are uncontrolled for other mpos and infinite bc.
+    trunc_par : dict
+        Truncation parameters. See :func:`truncate`
+
+    Returns
+    -------
+    mps : :class:`~tenpy.networks.mps.MPS`
```
jhauschild: call it `new_psi`

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+        MPS to apply operator on
+    mpo : :class:`~tenpy.networks.mpo.MPO`
+        MPO to apply. Usually one of make_UI() or make_UII(). The approximation being made are uncontrolled for other mpos and infinite bc.
+    trunc_par : dict
+        Truncation parameters. See :func:`truncate`
+
+    Returns
+    -------
+    mps : :class:`~tenpy.networks.mps.MPS` 
+        Resulting new MPS
```
jhauschild: Describe: 
```
Resulting new MPS representing `U_mpo |psi>`.
```

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+        #calculate good (rather arbitrary) guess for S[0] (no we don't like it either)
+        weight=np.ones(mpo.get_W(0).shape[mpo.get_W(0).get_leg_index('wL')])*0.05
+        weight[mpo.get_IdL(0)]=1
+        weight=weight/np.linalg.norm(weight)
+        S=[np.kron(psi.get_SL(0), weight)]
+    else:
+        S=[np.ones(Bs[0].shape[Bs[0].get_leg_index('vL')])]
+    #Wrong S values but will be calculated in mps_compress
+    for i in range(psi.L):
+        S.append(np.ones(Bs[i].shape[Bs[i].get_leg_index('vR')]))
```
jhauschild: `Bs[i].get_leg('vR').ind_len` is equivalent but shorter than `Bs[i].shape[Bs[i].get_leg_index('vR')]`
(Also 3 lines above for `'vL'` leg).

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    Parameters
+    ----------
+    i : int
+        First site.
+    mps : :class:`tenpy.networks.mps.MPS`
+        MPS to use on.
+    trunc_par : None|dict
+       If None no truncation is done. Else dict as in :func:`~tenpy.algorithms.truncation.truncate`.
+    """
+    theta=mps.get_theta(i)
```
jhauschild: explicitly add `n=2`, even if this is the default.

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    mps : :class:`~tenpy.networks.mps.MPS` 
+        Resulting new MPS
+    """
+    bc=psi.bc
+    if bc!=mpo.bc:
+        raise ValueError("Boundary conditions of MPS and MPO are not the same")
+    if psi.L != mpo.L:
+        raise ValueError("Length of MPS and MPO not the same")
+    Bs=[npc.tensordot(psi.get_B(i, form='B'), mpo.get_W(i), axes=('p', 'p*')) for i in range(psi.L)]
+    Bs[0]=npc.tensordot(psi.get_theta(0,1), mpo.get_W(0), axes=('p0', 'p*'))
```
jhauschild: If we simply use `new_mps = mps.MPS(..., form='B',...)`, we should *not* use the `theta` for the first `Bs[0]`.
In other words, delete this line...

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    for i in range(psi.L):
+        if i==0 and bc=='finite':
+            Bs[i]=Bs[i].take_slice(mpo.get_IdL(i) ,'wL')
+            Bs[i]=Bs[i].combine_legs(['wR','vR'], qconj=[-1])
+            Bs[i].ireplace_labels(['(wR.vR)'], ['vR'])
+            Bs[i].get_leg('vR').to_LegCharge
```
jhauschild: Missing brackets in the end. See comment below....

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+            Bs[i].ireplace_labels(['(wL.vL)', '(wR.vR)'], ['vL','vR'])
+            Bs[i].get_leg('vL').to_LegCharge()
+            Bs[i].get_leg('vR').to_LegCharge()
+
+    if bc=='infinite':
+        #calculate good (rather arbitrary) guess for S[0] (no we don't like it either)
+        weight=np.ones(mpo.get_W(0).shape[mpo.get_W(0).get_leg_index('wL')])*0.05
+        weight[mpo.get_IdL(0)]=1
+        weight=weight/np.linalg.norm(weight)
+        S=[np.kron(psi.get_SL(0), weight)]
```
jhauschild: I'm not sure if the order is correct, maybe it needs to be `np.kron(weight, psi.get_SL(0))`.
Please keep this comment for a moment...

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+            aR = range(W.shape[1]); aR.remove(IdL[i+1]); aR.remove(IdR[i+1])
+            
+            #Extract (A, B, C, D)
+            D = W[IdL[i] , IdR[i+1], :, : ]
+            C = W[IdL[i], aR, :, :]
+            B = W[aL, IdR[i+1], :, :]
+            A = W[aL, aR, :, :]
+    
+            WII = make_WII(dt, A, B, C, D)
+            #TODO from here on
```
jhauschild: I'll provide the correct charges shortly.
Just ignore this comment for a moment...

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+    IdL=H.IdL
+    IdR=H.IdR
+    v = np.empty_like(IdL)
+
+    c0 = np.zeros((1, H[0].num_q), dtype=np.int) #TODO 
+    U = []
+    for i in xrange(0, self.L):
+            W = H.get_W(i).itranspose(['wL','wR', 'p', 'p*'])
+            W = W.to_ndarray()
+            aL = range(W.shape[0]); aL.remove(IdL[i]); aL.remove(IdR[i]);
```
jhauschild: Might be more intuitive and actually faster to use boolean arrays:
```
aL = np.ones(W.shape[0], dtype=np.bool)
aL[IdL[i]] = aL[IdR[i] = False
```
Also, it might make more sense to name it `proj_L` instead of `aL`.
Same for `aR`, of course.

tenpy/algorithms/mps_compress.py
```diff
@@ -0,0 +1,345 @@
...
+        MPS to be compressed.
+        The singular values are usually ignored and need not be set to sensible values, except for bc='infinite', where psi.S[0] is needed as an initial guess.
+    trunc_par : dict
+        See :func:`~tenpy.algorithms.truncation.truncate`
+    """
+    bc=psi.bc
+    L=psi.L
+    if bc=='finite':
+        # Do QR starting from the left
+        B=psi.get_B(0,form=None)
```
jhauschild: Probably better to use the form explicitly...

tenpy/algorithms/mps_compress.py
scaletos: In the finite case the form is ignored in mps_compress. It's only assumed, that when contracting all tensors in the given form the correct state should be obtained. The leftmost Svalues are part of the state, shouldn't they be included? (In fact it's only cosmetic, as in finite case this is just one, but I think for finite the line is correct.

tenpy/algorithms/mps_compress.py
scaletos: As the leg was combined to `'wL.vL'` in that order, I think you're right. I change it for the moment and add a comment, that this is to be checked.

tenpy/algorithms/mps_compress.py
scaletos: The problem was, that the form was not correct anyway, as the contraction of MPO tensors and MPS tensors had no canonical form. The only thing I know and use is, that the full contraction in the given form results in the correct state without inserting any S values.

tenpy/algorithms/mps_compress.py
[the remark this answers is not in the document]
jhauschild: Still need to add a 
```
which : 'I' | 'II'
    selects the approximation made by choosing :func:`make_UI` (``'I'``) or :func:`make_UII` (``'II'``)
```

tenpy/algorithms/mps_compress.py
jhauschild: Yes. The problem I see, however, is that one might also use this function in other contexts than after `apply_mpo`, where the singular values are/might be correct. Then we should take proper care of the form to get the correct results....

scaletos: Then I could take the 'B' form here except for the first one in the finite case where I need the theta to get all S values. In apply_mpo I would then set exactly those forms, although it's wrong. Then this part works for a correct MPS.

tenpy/algorithms/mps_compress.py
jhauschild: That was a long moment.... Finally done.

jhauschild:
This pull request is merged into the `exp_MPO` branch inside this repository.
Some tests still don't work, but we'll continue working in the other branch (since we don't have write access to this branch)

PR #102, opened 2020-03-14
Feature/netcon

aneksteind:
I noticed you had a TODO to have netcon as a wrapper. I've created bindings for this for another project, so decided to add it here. this adds the ability to call netcon from python


aneksteind:
everything is passing now so it should be ready for review 👍

Issue #58, opened 2019-07-08
add_hc flag for MPO


jhauschild:
For models with e.g. Heisenberg couplings or fermionic/bosonic hoppings, explicitly adding the hermitian conjugate increases the bond dimension of the MPO. This increases the memory requirements of the `MPOEnvironment` during DMRG.

Instead of adding these terms explicitly to the MPO, we can set a flag to the MPO that the hermitian conjugate needs to be added later, and explicitly add the hermitian conjugate in the `matvec` operations of the effective OneSite/TwoSite Hamiltonians.

Steps: 
- [x] Add a flag to the CouplingModel that h.c. terms need to be added later.
- [x] Use this flag in the generation of bond terms of the NearestNeighbor model.
- [x] Add flag to MPO
- [x] Adjust generation of effective Hamiltonians to use alternative `matvec` depending on the flag.

Lschoon:
As an intermediate step:

- [ ] Add the flag to `add_coupling()` to create a h.c. term of the current coupling.
- [ ] Test this in at least one model.

Lschoon:
Question for @jhauschild: all documentation suggests that the hermitian conjugate of `add_coupling(t, u1, 'Cd', u2, 'C', dx)` is `add_coupling(np.conj(t), u2, 'C', u1, 'Cd', -dx)`. However, this is not followed in many model codes. Consider the following cases:

From `XXZChain`:
```
self.add_coupling(Jxx * 0.5, 0, 'Sp', 0, 'Sm', 1)
self.add_coupling(np.conj(Jxx * 0.5), 0, 'Sp', 0, 'Sm', -1)  # h.c.
```
Here, `u1 = u2 = 0`. The code as it is written here seems right; it creates a term `J S^+_i S^-_{i+1}` on the first line, then a term `J S^+_{i+1} S^-_i` on the second line. However, strictly following the documentation, the second line should be `self.add_coupling(np.conj(Jxx * 0.5), 0, 'Sm', 0, 'Sp', -1)`, which seems to give us a term `J S^-_{i+1} S^+_i`, i.e., the same term as the first line.

From `FermionModel` (similar example but with `u1 != u2`:
```
self.add_coupling(-J, u1, 'Cd', u2, 'C', dx)
self.add_coupling(np.conj(-J),` u2, 'Cd', u1, 'C', -dx)  # h.c.
```

From `SpinModel` (lines 1-2 disagree with documentation, lines 3-4 agree):
```
[...]

PR #104, opened 2020-03-26
add_hc and add_hc_to_MPO flags

Lschoon:
Solving https://github.com/tenpy/tenpy/issues/58 by introducing two flags: `add_hc` (for `add_coupling()` and `add_multi_coupling()`, to automate the inclusion of Hermitian conjugate terms) and `add_hc_to_MPO` (to reduce MPO bond dimension by not storing Hermitian conjugat terms, but computing these at runtime).


jhauschild:
Looks good. By the way, I think it we should start using branches inside the tenpy repository instead of keeping branches inside our private forked repositories. That would have allowed me to push changes to this pull request without asking you to merge changes from my repository.

Lschoon:
You're right, I will do that from now on. I need to be a little more diligent about pushing local commits, in that case ;)

Issue #49, opened 2019-07-06
Class for Paramters


jhauschild:
Currently, we store parameters as dictionary.
Make a subclass of the python build-in dictonary to replace the funcitonality of the functions in 
`tenpy.tools.params`:

- initiliaze from a dictionary (that was given to an algorithm, make a copy) and a `name` argument to be used for `get`. Set `verbose` as attribute.
- method for checking verbosity
- overwrite the `get` with functionality similar to `get_parameter`, taking only the key and default
- check for unused parameters by defining a `__del__` method.

jhauschild:
Maybe we should also think about how we document the individual parameters.
Currently, we have these huge tables. Maybe, it would be better to put them into a 
`Parameters` list?

Lschoon:
I'm happy to implement this (and have already started playing with it a little). Collecting some questions here:

- Do you definitely want this to be a subclass of the built-in python dictionary? It seems to me that all the desired functionality can be obtained by keeping the initial dictionary as an attribute of some generic class. There are many online sources warning me against subclassing `dict`. It also seems that there is some functionality in the built-in dict that we might not want (mainly accessing a value through `d['key']` without supplying a default). What are the expected benefits of subclassing `dict`? Perhaps a `collections.abc.MutableMapping` (see [documentation](https://docs.python.org/3/library/collections.abc.html)) is the solution.
- Should the `verbose` attribute be pulled from the initial dictionary (my preference), or be set through `__init__()`?
- Do you want to keep the `asarray` option for `get_parameter` somehow?
- I don't fully understand your comment. Yes, the parameter documentation is fairly bulky, but given the complexity of our algorithms that is not so surprising. How would you expect this `Parameters` list to be formatted?

Lschoon:
For context with my first question, the following code should implement most/all of the desired functionality, without any inheritance:
```python
class ParameterDict:
    def __init__(self, params, name):
        self.params = params
        self.verbose = params.get('verbose', 0)
        self.name = name
        self.unused = set(params.keys())

    def get_parameter(self, key, default):
        val = self.params.setdefault(key, default)  # get the value; set default if not existent
        self.print_if_verbose(key, default)
        self.unused.discard(key)  # (does nothing if already present)
        # if asarray:
        #     val = np.asarray(val)
        return val

    def print_if_verbose(self, key, default):
        name = self.name
        use_default = key not in self.params
        new_key = key in self.unused
[...]

PR #106, opened 2020-03-31
Parameter class

Lschoon:
See issue https://github.com/tenpy/tenpy/issues/49, this creates a class `Parameters` to collect parameter dictionaries and provide some useful functionality.


Lschoon:
Note that the build fails are currently due to an `ImportError` when trying to load `yaml`.

PR #118, opened 2020-12-02
Update example purification.py

anupam-mitra:
Import numpy to enable use of `numpy.sum` 
`numpy.sum` is used to calculate the z magnetization.


jhauschild:
good point, thank you!

PR #138, opened 2021-06-11
Fixed bugs in HelicalLattice

stefan-divic:
Fixed bug that was causing lat.order to return an incorrect list of lattice indices.


jhauschild:
Thanks for the pull request! 
I'll add .vscode to the gitignore ;-)

Issue #124, opened 2021-02-10
BUG: maximum recursion limit reached for large L models


jhauschild:
Generating a model with `L` larger than the python recursion limit fails. The default recursion limit for me is 3000, so it's not too severe, but I think it should be possible to avoid this, and it might even be more efficient...

**Minimal example**
```python
import tenpy
import sys
model_params = {'L': sys.getrecursionlimit() + 10}
tenpy.models.spins.SpinChain(model_params)
```
Reaches recursion limit in `travel_q_LR` defined inside `MPOGraph._calc_legcharges`.

PR #141, opened 2021-07-10
Fixing recursive issue with travel_q_LR as discussed in issue #124

sajantanand:
Changing `travel_q_LR()` in `mpo.py` to no longer be called recursively but instead run iteratively using a stack. This avoids exceeding the Python maximum recursion depth, as was occurring for larger systems.

### DIFF
--- a/tenpy/networks/mpo.py
+++ b/tenpy/networks/mpo.py
@@ -1631,14 +1631,13 @@ class MPOGraph:
         if infinite:
             charges[-1] = charges[0]  # bond is identical
         
-        def travel_q_LR_iter(i, keyL):
+        def travel_q_LR(i, keyL):
             """Transport charges from left to right through the MPO graph.
 
             Inspect graph edges on site `i` starting on the left with `keyL` and add charges
             for all connections to the right.
             Originally we recursively transported charges from there, but now this is done
             iteratively to avoid the maximum recursion limit in python for large systems."""
-            max_checks = 0  # How many items are ever entered into the queue
             stack = []
             stack.append((i, keyL))
             while len(stack):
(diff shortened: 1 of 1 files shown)


sajantanand:
I don't see that the tests have run or that they are successful.

tenpy/networks/mpo.py
```diff
@@ -1630,30 +1630,40 @@ def _calc_legcharges(self, Ws_qtotal):
...
+                    if qR is None:
+                        op_qtotal = site.get_op(ops[0][0]).qtotal
+                        ch_r[r] = qL_Wq + op_qtotal  # solve chargerule for q_right
+                        if infinite or i + 1 < L:
+                            edge_stack.append(((i + 1) % L, keyR))
+                            max_checks += 1
+                stack = edge_stack + stack
+            return max_checks
+        max_checks = travel_q_LR_iter(0, 'IdL')
+        print(max_checks)
```
jhauschild, line 1666: please remove this print.

tenpy/networks/mpo.py
```diff
@@ -1686,7 +1696,6 @@ def travel_q_RL(i, keyL):
         if not infinite and any([ch is None for ch in charges[-1]]):
             raise ValueError("can't determine all charges on the very right leg of the MPO!")
 
-        max_checks = sys.getrecursionlimit()  # I don't expect interactions with larger range...
```
jhauschild, line 1689: I think it actually makes more sense to hard-code `max_checks = 1000`. All of this is only to check for errors anyways; if you have a properly set-up MPO graph (e.g. from a `CouplingMPOModel`), it will terminate after just one iteration of the loop.
You can then also remove the corresponding code in `travel_q_LR_iter` that estimates `max_checks`.

jhauschild (changes requested):
Thanks for fixing this! Ready for merge as soon as you remove the print and hard-code max_checks.

jhauschild:
Closes #124

Issue #153, opened 2021-12-08
DMRG energy convergence criterium


rgbmrc:
Currently, if the energy increases with a DMRG sweep, the corresponding convergence criterium is always satisfied, **regardless of the magnitude of the increase**: 
https://github.com/tenpy/tenpy/blob/1bc2d4c221a2689b53392697494543ef12f738fe/tenpy/algorithms/dmrg.py#L832-L833
This is clearly stated in the documentation as well.

Especially in iDMRG, I sometimes observed a non-pathological large `Delta_E > 0`. Yet, I would not consider the optimization converged
unless `abs(Delta_E / max(E, 1.)) < max_E_err`. Is there some downside to using this condition instead of the current one?

jhauschild:
First of all: Happy new year, and thank you for all the pull requests! I'm glad to see contributions from outside :)

Finite DMRG is variational and should go down in energy, unless if you have a really strong truncation with wired edge cases.

Infinite DMRG is a bit more tricky, because you always just consider the energy in the bulk/central unit cell, and if you have non-uniform states, that can fluctuate; so it's more common to see the energy go up. Sometimes, it also makes a "jump" quite a bit up, bringing it down again afterwards. In most cases, I think it still approaches the energy from above, though, I'm not sure right now if I've ever seen a case where it comes from below.

I think the rationale behind the "stop if going up" is that we're making the state "worse" and "less close to the ground state", and we better stop before making it even worse. On the other hand, you're right that we're likely not completely converged in that case, and maybe should just be going. 

The other case where this criterion triggers, is when the truncation was too large compared to the DeltaE precision, say DeltaE = 1.-8, but E_trunc~=1.e-5. In this case, one often gets fluctuations of the energies on the order of E_trunc, and this criterion would stop the simulation.

I think the latter case is what we should worry about with the proposed fix - it doesn't make sense to continue to the full maximum number of sweeps if the truncation error and energy fluctuations induced by it are much larger than the DeltaE/DeltaS criteria. It would be much cleaner to separte this case into it's own criterion checking whether E_trunc_err > DeltaE, but it's a little bit tricky because E_trunc_err depends on the state as well - when exactly do we stop here? Maybe after a minimum number of steps?

rgbmrc:
Happy new year and thank you for your work on this amazing library!

In my case iDMRG was actually converging from below, as this energy VS sweeps plots show:

[...]

PR #154, opened 2021-12-09
Fix DMRG energy convergence criterium

rgbmrc:
If #153 is a bug, this should fix it.


Jakob-Unfried:
Have merged #326 instead

PR #156, opened 2021-12-26
Site: add operators in parallel

rgbmrc:
This PR implements a method `Site.add_ops` which parallelizes calls to `npc.Array.from_ndarray` when adding multiple onsite operators at once. `Site.__init__` and `.change_charge` now call this method instead of `.add_op`

The motivation for this change came from simulating a model with a large local Hilbert space. Its initialization took ~70 seconds, 95% spend in `from_ndarray`. While this is fine for production-stage simulations, it can be annoying during prototyping. With these changes, the wall time went down to ~10s on my machine.

Feel free to close without merging if you don't think this belongs to the base repository.

### DIFF
--- a/tenpy/networks/site.py
+++ b/tenpy/networks/site.py
@@ -8,10 +8,12 @@ import numpy as np
 import itertools
 import copy
 import warnings
+import multiprocessing
 
 from ..linalg import np_conserved as npc
 from ..tools.misc import inverse_permutation
 from ..tools.hdf5_io import Hdf5Exportable
+from ..tools.process import mkl_get_nthreads, omp_get_nthreads
 
 __all__ = [
     'Site', 'GroupedSite', 'group_sites', 'set_common_charges', 'multi_sites_combine_charges',
@@ -107,8 +109,7 @@ class Site(Hdf5Exportable):
         self.need_JW_string = set(['JW'])
         self.hc_ops = {}
         self.add_op('Id', npc.diag(1., self.leg), hc='Id')
-        for name, op in site_ops.items():
-            self.add_op(name, op)
+        self.add_ops(site_ops)
         if not hasattr(self, 'perm'):  # default permutation for the local states
             self.perm = np.arange(self.dim)
         if 'JW' not in self.opnames:
@@ -141,14 +142,16 @@ class Site(Hdf5Exportable):
             state_labels = self.state_labels.copy()
             for label in state_labels:
                 self.state_labels[label] = inv_perm[state_labels[label]]
+        ops = {}
         for opname in self.opnames.copy():
             op = self.get_op(opname).to_ndarray()
             self.opnames.remove(opname)
             delattr(self, opname)
             if permute is not None:
                 op = op[np.ix_(permute, permute)]
-            # need_JW and hc_ops are still set
-            self.add_op(opname, op, need_JW=False, hc=False)
+            ops[opname] = op
+        # need_JW and hc_ops are still set
+        self.add_ops(ops, hc=False)
         # done
 
     def test_sanity(self):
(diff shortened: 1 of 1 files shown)


jhauschild:
Hi, 
I'm a tiny bit worried by two things regarding this:
- Why is it necessary in the first place, i.e. why is the `from_ndarray` function so terribly slow in this case? Depending on the answer, that might be a hint that other parts in the code could also require some more optimization or might be unnecessary slow.
Your solution somehow feels a bit more like a workaround than of a "proper" optimization of that function (e.g. by having a Cython variant of it).
- The parallelization might be problematic if it uses more cores than a user requested on a computing cluster, which might even get the computing job killed. It's probably fine in most case, but I think that we might at least want to have some (config?) option to turn this on or off (maybe making this the default?) and alert the user that this might be a problem.

For the first point, I'd like to understand what's the issue. Would you mind sharing the local hilbert space dimension and/or local site in question?

rgbmrc:
Hi,

Concerning the second point, the number of processes is set using `mkl/omp_get_nthreads`. However, if OMP and MKL are both unavailable, the user has no way of controlling the number of processes used indeed.

Your first point is precisely the motivation for my last sentence in the PR description. I agree that this solution is really just a workaround; even though a very effective one, at least for my purposes. 

The local Hilbert space is 54-dimensional and has 21 operators defined (I plan to include more). A unit cell can be 2 or 4 sties.

### Optimizations

Making other parts of TeNPy's code more efficient might involve not rebuilding the npc operators from sratch (i.e. from ndarrays) in `change_charge` (?). However this does not look trivial. 

On my side, hopefully I already removed spurious  `from_ndarray` calls by not sorting the leg during site initialization, since `set_common_charges` takes care of that during lattice initialization. The function is still called twice (for every operator, for each site of the unit cell): by `Site.__init__` and by `set_common_charges`. The first call can be avoided adding the operators during lattice initialization, but this clutters the code and results in a site class with no operators. I fear that further optimizations require changes to the code that produces the input data for the site, and I would rather avoid those.

### Outlook

I see the following - not necessarily mutually exclusive - possibilities:

1. Ignore (I am using my fork anyway), at least unless other users face the same problem.
2. Cython `from_ndarray`.
3. A `.tools.process._ncpus` attribute with get/set functions, related or unrelated to OMP & MKL nthreads (if available) but not relying only on those; handy if multiprocessing is introduced in other regions of TeNPy's code.
4. A config option for `Site`, (does this really look like a user-config-option? also currently `Site` takes no config and I noticed that typically configs are mandatory arguments); or ...
5. ... a `parallelize_add_ops=False` optional argument in `Site`'s constructor (looks a bit ad-hoc).

If you feel this PR is worth including I can implement 4. or 5. Due to time constraints that's the best I can do right now, but if you favour other solutions I might be able to come back to this in the future :)

Jakob-Unfried:
I was playing around with optimizing `from_ndarray` for this purpose.
I could not replicate it being anywhere near as slow as you report.
For a local operator with a O(50) diimensional leg i get runtimes of about a microsecond for a single `from_ndarray` call, no matter how complicated or weird i make the charges.

@rgbmrc Can you share a minimal code that exhibits the slow `from_ndarray` codes?
Mostly looking for your leg charge and the operators you are adding

rgbmrc:
Hi @Jakob-Unfried, I tested this again with an updated tenpy installation and the cumulative time spent in `from_ndarray` and subcalls was now around 20s (0.1s per call), which is not a big deal but is still slightly cumbersome. You can find the code I used at [rgbmrc/tenpy-add-op-test](https://github.com/rgbmrc/tenpy-add-op-test). The proper MWE is the `LGT_unit_cell` function. I still included the whole model initialization (with unit parameters) to show that 90% of the time is spent inside `from_ndarray`. 

Physical legs have four Z_3 and two U(1) charges. Operators are from [this group repository](https://baltig.infn.it/qpd/qhlgt-models), included as a submodule. You can find them in '' load them via `scipy.sparse.load_npz`. `from_ndarray` is called 188 times = 23 [num ops] * 4 [Lu] * 2 [due to set_common_charges] (the last factor can be avoided but imho it is not worth to clutter the code for it).

Jakob-Unfried:
I get about 30% performance increase in `from_ndarray` calls from the changes in #356.
The changes there are in summary to not check the blocks that are forbidden by the charge rule individually during the loop and instead only check the remaining entries once after the loop.

That was the first low-hanging fruit that @jhauschild and I found.
The next level would be to cythonize.

@rgbmrc For your usecase, if you are storing the operators on disc anyway, why not store tenpy arrays? e.g. vie the HDF5 interface?

PR #159, opened 2022-01-26
Add norm_tol_final parameter to DMRGEngine

### LINKED ISSUE #158: DMRG norm_tol convergence criterion
Jack pointed out to me that there is an inconsistency in the `norm_tol` parameter of DMRG:

> In DMRGEngine._canonicalize, the default norm_tol is 1e-5, while the test for whether an MPS is canonical or not in Tenpy seems to be 1e-10 (e.g. in MPOTransferMatrix .__init__). This means that one cannot, for example, naively use the output of DMRGEngine as the initial state for a new ground_state_search simulation from file without risking a "not in canonical form" error. Of course, the problem is easily solved for a finite MPS -- one can just call .canonical_form after DMRG -- while for iMPS we can just set the norm_tol to 1e-10 appropriately and possibly also raise norm_tol_iter. I guess my question is, why are these values for the norm tolerance so different and which should I prefer for iDMRG simulations? It naively seems to me that setting the norm_tol to 1e-5 might lead to an MPS quite far from canonical form. 

That's not at all a naive question, and it raises an issue in about the DMRG convergence criteria in TenPy....

The norm_tol is used during DMRG, where one *expects* that the state is not yet in canonical form until DMRG fully converges. Especially for infinite systems with long correlation lengths, this convergence is sometimes quite slow, so we allowed a very moderate value of 1.e-5 just to get DMRG to converge at all for these cases.

On the other hand, I've required a quite good precision for the MPOTransferMatrix because I found that any error of the MPS not being in canonical form propagates to wrong environments, which can screw up *all* of the following DMRG run - and one can simply circumvent this by calling `psi.canonical_form()` before initializing the MPOTransferMatrix.

Calling psi.canonical_form is
- *bad* if you want to re-use the environments, because it renders the environments invalid - so we try to avoid it whenever one might need to reuse the environments
- but *good* whenever you want to re-use psi in another way: expectation values, MPOTransferMatrix etc assume canonical form, and any error in the canonical form propagates to wrong expectation values.

Historically, we only added the MPOTransferMatrix with the ability to recover the environments from psi in version 0.9.0 last summer.
This allows us to be a bit more on the "safe" side, and explicitly call psi.canonical_form at the end of DMRG, and rather re-generate environments for the next DMRG run, if necessary. However, right now, TeNPy is still using the "old" mode trying to avoid rendering the environments invalid...
Given that we have the MPOTransferMatrix now, I think we should add "split" the norm_tol parameter up into one "norm_tol" to be acchieved for DMRG convergence, and one `norm_tol_final` to be used for the state returned by DMRG after convergence by calling psi.canonical_form, even if this renders environments invalid. The `norm_tol_final` could then be significantly more precise than the `norm_tol` for convergence...

Jack-Kemp:
Fixes #158.

First attempt at adding the `norm_tol_final` parameter to ensure canonicalization of the output of DMRGEngine even when the `norm_tol` parameter is less strict due to possible convergence issues (see discussion in #158).

**Currently untested.** Comments on implementation welcome, as are ideas for a simple test of the norm convergence to put in the tests (currently there doesn't seem to be a test of the norm_tol parameter and norm_tol_iter?). This requires a simple case where DMRG struggles on the norm: perhaps transverse field Ising at the critical point J=g is good enough?

### DIFF
--- a/tenpy/algorithms/dmrg.py
+++ b/tenpy/algorithms/dmrg.py
@@ -766,8 +766,12 @@ class DMRGEngine(Sweep):
             norm_tol_iter : float
                 Perform at most `norm_tol_iter`*`update_env` sweeps to
                 converge the norm error below `norm_tol`.
-                If the state is not converged after that, call
-                :meth:`~tenpy.networks.mps.canonical_form` instead.
+            norm_tol_final : float
+                After performing `norm_tol_iter`*`update_env` sweeps, if
+                ``np.linalg.norm(psi.norm_err()) < norm_tol_final``, call
+                :meth:`~tenpy.networks.mps.canonical_form` to canonicalise
+                instead. This tolerance should be stricter than `norm_tol`
+                to ensure canonical form even if DMRG cannot fully converge.
             P_tol_to_trunc : float
                 It's reasonable to choose the Lanczos convergence criteria
                 ``'P_tol'`` not many magnitudes lower than the current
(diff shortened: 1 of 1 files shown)


jhauschild:
Looks perfect, thank you!
I quickly checked that it works properly, and that all tests pass, so I'll just go ahead and merge this.

PR #168, opened 2022-05-26
added the `NLegLadder` lattice coupling N chains

nheinsdorf:
I generalized the ladder class to couple N chains. 

As next-Nearest-Neighbors I put the sites in the neighboring unit cell that carry the same sublattice index. If the unit cell has length a=1 in both x- and y-direction, obviously the sites in the same unit cell are closer (or equally close for N=2) than hopping to the next unit cell. However, I found this the most intuitive labeling. Same for the next-next-Nearest-Neighbors that I set to be the sites from the next unit cell with the sublattice index changed by one.  
If you think these labels will lead to confusion, I will just remove the nNN and nnNN (and maybe also NN) attributes from the new class.

### DIFF
--- a/tenpy/models/lattice.py
+++ b/tenpy/models/lattice.py
@@ -2236,6 +2236,56 @@ class Ladder(Lattice):
                 assert (False)  # should not be possible
             return ordering
         return super().ordering(order)
+      
+      
+class NLegLadder(Lattice):
+    """A ladder coupling N chains.
+
+    .. plot ::
+
+        import matplotlib.pyplot as plt
+        from tenpy.models import lattice
+        plt.figure(figsize=(5, 1.4))
+        ax = plt.gca()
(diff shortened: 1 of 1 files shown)


jhauschild:
Cool, that looks awesome, thanks for the pull request!!!
It's always great to see someone actually contributing, without asking for it :)
As you can see, I've already merged the pull request. 

Yes, I think the labeling of pairs as (next...) nearest neighbors is confusing. The pairs can actually have arbitrary names. 
I renamed them to `rung_NN`, `leg_NN`, and `diagonal`, does that sound reasonable to you?
For consistency, I added the same names to the (two-leg) `Ladder` as well.
I kept `nearest_neighbors` as the union of `rung_NN` and `leg_NN`.

Also, I've implemented the folded order.

PR #169, opened 2022-06-17
Modified the convergence judgement function of lunczos and Arnoldi, s…

Eta-Chu:
…pecifically the indexing error when calculate the RitzRes.


jhauschild:
As I already wrote in the email:  Respect, I'm impressed that you spotted this! Thanks for the pull request! :+1:

PR #177, opened 2022-09-09
Vumps

sajantanand:
Starting the VUMPS pull request. This is not that close to being merged, as I realize there is a lot more changed in this branch than I realized. We've implemented single-site PW, half of multi-site PW, half of domain wall PW, and all the segment stuff we've done so far (so things work for U(1) but not Zn). I'll just start the PR so we can keep track of things that need to be fixed.


sajantanand:
Since we aren't chatting for a week, I'll finalize multi-site PWE by then so that everything but the DW excitations with ZN charges can be pushed.

jhauschild:
current state of the vumps branch:
- [x] VUMPS single site
- [x] VUMPS two-site
- [ ] VUMPS with mixer
- [x] GMRES
- [ ] Switch from arnoldi to  GMRES for generating MPO environments
- [ ] fix Zn charges for segment Domain wall excitations
- [x] single site plane wave
- [ ] multi-site plane wave (exponential scaling partially implemented)
- [ ] mulit-site plane wave (DMRG-style with better scaling)
- [ ] topological versions for all plane waves
- [ ] getting environments in parallel (maybe? might have bad communication overhead....)

PR #179, opened 2022-10-07
bugfix MPS.gauge_total_charge

johannes-moegerle:
This is a short bugfix for gauging the total charge of a MPS.

The addressed bug mainly introduced errors when trying to calculate overlaps of two MPS with different _outer_virtual_legs (e.g. after applying a spatial inversion to one MPS with total charge != 0).

The following code is a minimal example to reproduce the bug.
```
from tenpy.networks.mps import MPS
from tenpy.networks.site import SpinSite

p_state = 1*['up'] + 10*[1] + 1*['up']
sites = [SpinSite(S=1, conserve='Sz')]*len(p_state)
psi = MPS.from_product_state(sites, p_state)

other = psi.copy()
psi.spatial_inversion()
# print(f"Before overlap: {psi._outer_virtual_legs()=}")
overlap = psi.overlap(other)
# print(f"After overlap: {psi._outer_virtual_legs()=}")
print(f"Successfully calculated the overlap: {overlap}")

# Note that with the bugfix the virtual legs of psi are getting changed when psi.overlap is called.
```

Additional question: Should the docstring in MPS.overlap mention, that the function might change the input MPS (or rather its virtual legs)? Or should it make a copy instead of silently changing the virtual leg charges?

### DIFF
--- a/tenpy/networks/mps.py
+++ b/tenpy/networks/mps.py
@@ -1529,7 +1529,7 @@ class MPS:
             if np.any(vL_chdiff != 0):
                 # adjust left leg
                 self._B[0] = B.gauge_total_charge('vL', B.qtotal + vL_chdiff, vL_leg.qconj)
-            B.get_leg('vL').test_equal(vL_leg)
+            self._B[0].get_leg('vL').test_equal(vL_leg)
         for i in range(self.L):
             B = self._B[i]
             desired_qtotal = qtotal[i]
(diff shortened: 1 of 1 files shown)


jhauschild:
Thanks a lot for the fix and the suggestions I've made a small follow-up commit which
- adds a comment in `gauge_total_charge` that it's modifying the MPS in place, and
- makes a shallow copy in MPS._gauge_compatible_vL_vR() before calling gauge_compatible_charge.

PR #182, opened 2022-10-17
Measurement function fixes 

jhauschild:
Again, this change is backwards incompatible, so I'm creating a pull request to make the change more visible and track it better.

The relevant incompatibility is in f58050620ddd9282c4e99a820517d49935363462 (cherry-picked from 2a5052ac2abc672b0f2e83d8f74fb3ca3be2229b), namely that measurement functions need to accept models as additional arguments. This is necessary to ensure compatiblity of the `psi` to be measured with the model, e.g. when grouping sites.

While you're changing those measurement functions, you might also want to follow the new TeNPy convention that they start with an `m_`, e.g. we renamed `tenpy.simulations.measurement.bond_dimension` to `tenpy.simulations.measurement.m_bond_dimension`.  For now, the old names still work in the config/yaml files, but they give a FutureWarning. 
You should update the config yaml files in projects where you want to update to a new TeNPy version:
- update the `connect_measurements` simulation parameter: all direct measurement functions (taking `results, psi, model, simulation` as arguments and saving to `results`) defined in TeNPy have been renamed to start with `m_`.
  At this point, all those functions are defined in `tenpy.simulations.measurement.py`.
- If you're using `psi_method` to measure e.g. correlation functions, you need to update the entry as follows:
An old python entry for connect_measurements
```python
{...,
 'connect_measurements': [['tenpy.simulations.measurement',
                           'psi_method',
                           'correlation_function',
                           {'key': '<Sp_i Sm_j>', 'ops1': 'Sp', 'ops2': 'Sm'},
                          ],
                          ...
                          ]
```
can get replaced with new entry:
```python
{...,
 'connect_measurements': [['psi_method',
                           'wrap correlation_function',
                           {'key': '<Sp_i Sm_j>', 'ops1': 'Sp', 'ops2': 'Sm'},
                          ],
                          ...
                          ]
```

Similarly, an old yaml entry for connect_measurements
```yaml
connect_measurements:
    - - tenpy.simulations.measurement
      - psi_method
      - method: correlation_function
        key: '<Sp_i Sm_j>'
        ops1: Sp
        ops2: Sm
    - ...
```
can get replaced with new yaml:
```yaml
connect_measurements:
    - - psi_method
      - wrap correlation_function
      - results_key: '<Sp_i Sm_j>'
        ops1: Sp
        ops2: Sm
    - ...
```

### DIFF
--- a/doc/changelog/latest.rst
+++ b/doc/changelog/latest.rst
@@ -5,6 +5,9 @@ Release Notes
 -------------
 Backwards-incompatible rewrite of TDVP!
 
+Note that measurment functions for simulations need to be updated to accept another `model` parameter
+
+
 Changelog
 ---------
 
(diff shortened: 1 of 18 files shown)


jhauschild:
Further information can also be found in the new [measurement user guide](https://tenpy.readthedocs.io/en/latest/intro/measurements.html)

PR #183, opened 2022-10-20
Shift symmetry

jhauschild:
Creating a pull request in an effort to finally merge this.

What's actually missing for this?

- [x] add a dipole-conserving example model using the shift symmetry
- [x] Write a small test to ensure everything works with non-trivial shift symmetry
- [x] Document the shift symmetry in the user guide


Jakob-Unfried:
I will make an effort to merge this. The notes below are mostly for my own benefit and FYI for you guys

- Having the `_shift_func` as part of `__getstate__` seems like a bad idea. It would probably cause problems when unpickling charges...
Possible alternatives:
  - let shifting be a method of the Site or of the Model
  - if it is possible to parametrize all shifts that we can anticipate, pickle those parameters instead

- Double check any direct access to `MPS._B`, `MPS._S` and `MPO._W` (should probably only access via `get_B` etc, to allow charge shift to intercept)

- Double check site indices used in `MPS.get_B`, `MPS.get_S` and `MPO.get_W`: should *not* cast `_to_valid_index` before!

jhauschild:
Just saying, it's probably a good idea to resolve the merge conflicts and merge the current main  branch first, since #267 moved around measurmeent functions in mps.py as well, which this PR partially had to update as well. Let me know if you need help with that.

Jakob-Unfried:
The merge introduced an error with purification MPS.
I have to stop working on this for now, so i pushed the mostly-complete merge

Jakob-Unfried:
@phzechm Do you have code for dipole conserving (or other) models using the shift symmetry?

jhauschild:
I agree to your proposal. In fact, recently I stumbled across code with Philipp Thoma where -1 was used for the last site for finite bc, and was initially surprised that we had allowed that.
Yet, raising the error when it was allowed previously is a "strongly" backwards incompatible change that might surprise users and breaks existing legitimately working code (since negative indices are a commonly used pattern in python...)
I think we should for now just issue a warning, and only introduce the error from version 0.99 to 1.0.
However, any code inside tenpy should already now be fixed to not use negative indices (by checking that the tests run without warnings).

Jakob-Unfried:
I have implemented the warning on negative indices for open boundary conditions in `_to_valid_index`.
When i tried to make sure that tenpy internal code does not issue that warning, i found that it does occur in the TDVP tests.
In particular in `test_tdvp.py::test_tdvp` and `test_time_evolution.py::`

The negative indices come from `Sweep._cache_optimize`. In particular in the `self.n_optimize == 2` and `not move_right` branches, the `kwargs['preload_LP'] = i0 - 1` is sometimes negative.
In particular we have `i0=0` but `move_right=False` when the negative index occurs for the first time.
Is this a bug in the TDVP sweep schedule?

Jakob-Unfried:
I am fixing the doctest action btw, see #291

Jakob-Unfried:
Any good ideas how to deal with `MPS.spatial_inversion()`?
Would it be correct to just shift the charges according to how much a tensor has moved?
I.e. can we do sth like
```python
new_Bs = [old_Bs[L - i].replace_labels(...).transpose(...).shift_charges(L - 2 * i) 
          for i in range(L)]
```
?

jhauschild:
No, spatial inversion won't work like this, at least for infinite systems.
The dipole charge is something like (sum_x x * n_x), which under inversion goes to (sum_x (-x) n_x). So you would need flip all those charges and define a new corresponding `shift_charges`.

Jakob-Unfried:
Two more questions, while I have your attention here ;)

1. The current `ChargeInfo._shift_func` attribute, which is an arbitrary function defined
    elsewhere probably causes problems with pickling. I think it should be a class method of
    some class that is specialized to deal with dipolar charges. (Of course this allows other
    non-translationally-invariant symmetries to be defined in the same way).
    
    a.  This could be the `ChargeInfo`, i.e.
    ```python
    class ChargeInfo:
      ...
      def _shift_func(self, charges, shift):
        return charges
    
    class DipolarChargeInfo(ChargeInfo):
      charge_idx = 0  # idx of the "charge"
      dipole_idx = 1  # idx of the "dipole moment of charge_idx"
      def _shift_func(self, charges, shift):
        charges[:, self.dipole_idx] += shift * charges[:, self.charge_idx]
    ```

    b. Or in `Site`
    ```python
    class Site:
      ...
      def shift_charges(self, charges, shift):
        return charges
    
    class DipolarSpinSite(Site):
      ...
      def shift_charges(self, charges, shift):
        charges[:, 1] += shift * charges[:, 0]
    ```
  I think that b is a bit cleaner, since we dont need extra classes (`DipolarSpinSite` or similar
  needs to be defined either way), and whenever we have an MPS we also have access to the site.

2. With the mixer, where the `S` carry charges, for infinite MPS, we should probably set `_S[L]`
to be a *shifted* version of `_S[0]`, no longer the same. Does this cause any problems elsewhere?

Jakob-Unfried:
Another issue i just thought of:

3. The current way of implementing the shift only works with regular chains. For a general lattice, the meaning of `shift`, which is distance along the MPS, depends on the lattice and the position of the site before shifting.
Think e.g. of a two-rung ladder, for every other site along the MPS, shifting should not change the dipole "charge".
If we want to implement the dipole symmetry in such a way that it can be eventually used for all lattices, we need to define the shifting functions such that they have access to the lattice, which makes the interface more complicated.
Should we do this now?

jhauschild:
1. Functions can be pickled and saved with `hdf5_io`, if they are a) global in some module that can be imported at load time and don't have a local scope that needs to be saved, or b) if they are not part of the data to be saved, because they are (class)methods of a (global) class that can be imported at load time, and you only safe instance data like the `__dict__` (which doesn't contain class methods that get bound to the instance). If you explicitly save a reference to a method, e.g. by returning it in the __getstate__, it will still work: pickle will internally save the state of the instance, I think, and hdf5_io will (currently) fall back to the pickle method raise a warning.
That said, even if that would work in theory, it's certainly a good idea to avoid it, as it could lead to a lot of headaches later on...

I also like the thinking ahead of Jakobs point 3. 
Really, the information we need is the lattice.mps2lat_idx, which depends on the order and dimension of the lattice.
Also, we need to replace the single `shift` arguments by a pair `shift_from` and `shift_to`, which we can translate into dx with the `lattice.mps2lat_idx`. We might even want to take into account the lattice `basis` and `position`, but I guess that's a different story.

Intuitively, I was leaning towards b) as well. However, the "cleanest" solution would definitely be a), since in general we could have two independent sites in a models, one of which has a diploe charge, while the other is say just a spin without electric charge. That case would be very awkward in solution b), while for a) they naturally just share the same ChargeInfo (subclass).

For completeness, let me propose another solution c), which would be to subclass the MPS with methods for the shift.
That subclass could exceptionally get a reference to the lattice as well, while the current MPS is agnostic of it.
The drawback of c) would be that it is a bit counterintuitive to setup, and we don't have access to the functions e.g. from the MPO! 
The advantage of c) would be that we can would avoid cyclic references  `lattice.unit_cell -> site, DipolarSpinSite.leg.chinfo -> ChargeInfo -> lattice.mps2lat_idx` that we would get in solutions a) and b).
However, I think that those cyclic references actually aren't too bad: python memory management/pickle/hdf5_io can handle such things, and they just interlock data that we anyways have around in memory during the whole simulation and routinely save together in the output files anyways.

For those reasons, I think the cleanest solution would be a).

2. Agreed. I think it should be fine if you do it consistently, you can probably just update the `set_SL/SR` method.
3. See 1 above.

Jakob-Unfried:
I have lots of unpushed changes on this. If anyone starts contributing to this, please coordinate with me first.

Jakob-Unfried:
Loose ends:
- Should test the upcoming VUMPS for how it interacts with shift-symmetry. If it doesnt, raise a warning.
- [ ] Should we adjust `m_correlation_length`? In partiuclar, using `lat.mps_unit_cell_width` instead of `lat.N_rings` should be more stable against grouping sites.
- [ ] merge the companion PRs in the submodules
- [ ] bump the submodules
- [ ] the tests with the dipole conserving spin chain on an infinite system seem pretty unstabel... investigate ``tests/test_time_evolution.py::test_ExpMPOEvolution_dipolar`` and ``tests/test_dmrg.py::test_dmrg_dipole_conservation``. Is this to unstable to merge...?

Jakob-Unfried (approved):

Jakob-Unfried:
Only thing missing now is to adjust the new MPO environment building from #507 .

I am happy with doing only the bare minimum to get iDMRG running again and deferring edge cases / clean up to #560.
We should merge this ASAP, so we can get started with v2.

I dont understand the details of the new MPO environment builder, and hopefully we can fix this without me having to understand the details.
I have asked @anton-romen for help...

Issue #180, opened 2022-10-08
Undefined varaible in ground_state_search.OrthogonalExcitations (?)


yuchihe:
see "self.results['ground_state_energy'] = E0"

E0 seems to be undefined.

Jakob-Unfried:
It seems like [this commit](https://github.com/tenpy/tenpy/commit/64ee7a41827f0cf23f19b2cf363d7d8b322fcc67#diff-05de523b29605437841b7c95288a4feca15ca962a088353da3fe76f6f33ab83fR175) introduced the bug

jhauschild:
actually, this was solved in another branch, excitation-science, see also #178.

The real issue is that the branches became a mess, and I need to sort out which branch contains what on top of what.... I'll try to do that asap, should have done that much earlier!

PR #198, opened 2023-02-06
fix issue 180: quick and dirty version

Jakob-Unfried:
This should fix issue #180.
I did not have the time, however, to understand what the "optimization" was that the comment references.
Apparently, the original idea was to evaluate the groundstate energy later.
My fix ignores this and computes it right away

@jhauschild do you remember what this was about?


jhauschild:
getting the MPO environments is expensive for large MPO bond dimensions, that's why I wanted to optimize it to only once get them. The corresponding optimization is implemented in the excitation_science -> excitation_science_topo_refactor -> vumps branches.

Jakob-Unfried:
Closing this, since we will fix the issue by merging those branches.

PR #199, opened 2023-02-07
Refactor npc

Jakob-Unfried:
## Goal
- Allow the numerical backend to be swapped out with minimal or no changes to algorithm codes.
- Provide these numerical backends

For example, a DMRG simulation code should be the same, no matter if symmetries are exploited or not or if the numerics is performed on CPU via numpy or on GPU via CUDA.

## Rough Roadmap

- [x] Create a modular interface that allows for different backends
- [x] Implement, test a numpy backend: numpy blocks, no symmetries
- [x] Implement, test a torch backend; torch blocks, no symmetries
- [x] Implement abelian backend in pure python
- [x] BlockBackend for generic library which follows the [Array API](https://data-apis.org/array-api/latest)
- [ ] Write extensive tests for `tenpy.linalg`
- [ ] Documentation
- [ ] Review interface
- [ ] Integrate into rest of tenpy
- [ ] Write our own CUDA support without torch?
- [ ] Optimization for abelian backend, e.g. C-extension?
- [ ] Implement a non-abelian backend

## Loose ends for later
- [ ] in-place operations? mutability? When do we need to make copies / view of data? of symmetry/label/... metadata?
- [ ] I/O support for tensors (`__getstate__` and `__setstate__` for pickle and hdf5 interface)
  - For e.g. torch backend, make sure the stored data is portable, e.g. does not require a GPU to load...
    Pickling a `torch.Tensor` that is on the GPU can not trivially unpickled on a GPU-less machine.
  - Backend deserialization needs to execute initialization code (such as e.g. importing torch)!
- [ ] Scalar class?
- [ ] Could re-introduce `VectorSpace.__mul__` in an associative way, if `ProductSpaces * VectorSpace` gives a *flat* `ProductSpace` instead of nesting
- [ ] optimization: when to perform which input checks
- [ ] think about real vs complex numbers, what it means for symmetries, are singular values "real", dtype, ...
- [ ] For the groups and their sectors I (Jakob) have assumed that we consider complex vectorspaces. Is the real representation theory substantially different for any group that we support?
- [ ] how to support seeds for random generation?
- [ ] currently using only a single device for torch
- [ ] `Tensor.__del__` should free up all memory. default is probably ok, but think about it again!
- [ ] Make sure concepts have unique names (e.g. `charge, sector, irrep` ; `dual, conj` ; `leg, space, axis` ; `...?`
- [ ] Error handling:
  - [ ] Make error messages informative
- [ ] documentation for the two inequivalent ways of simulating fermions (JW in the operators vs FermionParity symmetry)
- [ ] Double-check the entire `test_sanity` call chain


Jakob-Unfried:
Closing this in favor of #309

PR #200, opened 2023-02-19
Use and test for correct year in copyright notices

Jakob-Unfried:


Jakob-Unfried:
@jhauschild you ok with this?

PR #202, opened 2023-02-24
Qr based tebd

Jakob-Unfried:


Jakob-Unfried:
@jhauschild 
Even [`if not expand`](https://github.com/tenpy/tenpy/blob/281e929c1b0f43fba48d1da3e69caa76e4deb3ce/tenpy/algorithms/tebd.py#L617)
we should still call `svd_theta`, because `self.trunc_params` could have a `svd_min` that needs to be respected, right?

PR #203, opened 2023-03-02
update QRBasedTEBDEngine

nikirchner:
Small update for the QR-based TEBD algorithm.
`update_bond_imag` is now implemented if we allow for SVD; it is not yet working using eigh but there is some code behind an error message that might be useful for the implementation using eigh.


Jakob-Unfried:
thanks =)

PR #205, opened 2023-03-08
Qr based tebd

Jakob-Unfried:
## TODO before merging:
- [x] Properly extract truncation error! Currently, only the error of the SVD of `Xi` is returned, while errors introduced by `_qr_based_svd` are ignored
- [x] Are there easy diagnostics that we can run to detect that `expand` was too small?
  - no good ideas for now, writing a ``.. todo ::`` for now
- [x] Even [if not expand](https://github.com/tenpy/tenpy/blob/281e929c1b0f43fba48d1da3e69caa76e4deb3ce/tenpy/algorithms/tebd.py#L617) we should still call svd_theta, because self.trunc_params could have a svd_min that needs to be respected, right?
- [x] We probably want more elaborate control over `expand`. E.g. even if a moderate expansion ratio of 10% is eventually appropriate, it may be prohibitive in the beginning, e.g. for the very first time-step starting from a product state.
- [x] Make `expand` an option in the config, not an arg of `update_bond`, same for `use_eig_based_svd`
- [x] docstring

### DIFF
--- a/tenpy/algorithms/tebd.py
+++ b/tenpy/algorithms/tebd.py
@@ -723,7 +723,7 @@ class QRBasedTEBDEngine(TEBDEngine):
                 s_new = min(s_new, sizes_new[j_new])  # don't go beyond block
                 start = vL_new.slices[j_new]
                 piv[start:start+s_new] = True
-            #  Y0.iproject(piv, 'vL')
+            Y0.iproject(piv, 'vL')
             Y0 = Y0.ireplace_label('p1', 'p').iconj()
         else:
             Y0 = self.psi.get_B(i1).conj()
--- a/tenpy/algorithms/tebd.py
+++ b/tenpy/algorithms/tebd.py
@@ -591,6 +591,58 @@ class Engine(TEBDEngine):
         TEBDEngine.__init__(self, psi, model, options)
 
 
+class QRBasedTEBDEngine(TEBDEngine):
+    """Similar as standard TEBD, but based on QR decompositions only.
+
+    """
+    def update_bond(self, i, U_bond):
+        i0, i1 = i - 1, i
+        logger.debug("Update sites (%d, %d)", i0, i1)
+        # Construct the theta matrix
+        C = self.psi.get_theta(i0, n=2, formL=0.)  # the two B without the S on the left
+        C = npc.tensordot(U_bond, C, axes=(['p0*', 'p1*'], ['p0', 'p1']))  # apply U
+        C.itranspose(['vL', 'p0', 'p1', 'vR'])
+        SL = self.psi.get_SL(i0)
+        if isinstance(SL, npc.Array):
+            theta = npc.tensordot(SL, C, axes=['vR', 'vL'])
+        else:
+            theta = C.scale_axis(SL, 'vL')
+
+        # here start differences to normal TEBD
+        Y0 = self.psi.get_B(i1).conj()
+        # TODO: choose Y0 as slice of theta
+
+
+        theta_i0 = npc.tensordot(theta, Y0,
+                                 axes=[['p1', 'vR'], ['p*', 'vR*']]).ireplace_label('vL*', 'vR')
+        theta_i0 = theta_i0.combine_legs(['vL', 'p0'], qconj=+1)
+        A_L, Xi = npc.qr(theta_i0, inner_labels=['vR', 'vL'])
+        A_L = A_L.split_legs(['(vL.p0)'])
+        theta_i1 = npc.tensordot(A_L.conj(), theta,
+                                 axes=[['vL*', 'p0*'], ['vL', 'p0']]).ireplace_label('vR*', 'vL')
+        theta_i1 = theta_i1.combine_legs(['p1', 'vR'], qconj=-1, new_axes=0)
+        theta_i1.itranspose(['(p1.vR)', 'vL'])
+        B_R, Xi = npc.qr(theta_i1, inner_labels=['vL', 'vR'], inner_qconj=-1)
+        B_R = B_R.split_legs(['(p1.vR)']).replace_label('p1', 'p')
+        Xi.itranspose(['vL', 'vR'])
+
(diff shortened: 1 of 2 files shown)


Jakob-Unfried:
Even [`if not expand`](https://github.com/tenpy/tenpy/blob/281e929c1b0f43fba48d1da3e69caa76e4deb3ce/tenpy/algorithms/tebd.py#L617)
we should still call `svd_theta`, because `self.trunc_params` could have a `svd_min` that needs to be respected, right?

Jakob-Unfried:
Ready to merge in my opinion.
@jhauschild ?

tenpy/algorithms/tebd.py
```diff
@@ -591,6 +591,264 @@ def __init__(self, psi, model, options):
...
+        To use `use_eig_based_svd == True`, which makes sense on GPU only, we need to implement
+        the `_eig_based_svd` for "non-square" matrices.
+        This means that :math:`M^{\dagger} M` and :math:`M M^{\dagger}` dont have the same size,
+        and we need to disregard those eigenvectors of the larger one, that have eigenvalue zero,
+        since we dont have corresponding eigenvalues of the smaller one.
+
+    .. todo ::
+        Can we implement some simple diagnostics that issue a warning if the CBE expansion
+        was too small?
+        Could potentially look into the tail of the schmidt spectrum.
```
jhauschild, line 609: I don't think that's necessary. Are you afraid that the expansion factor was too small? It can grow exponentially with the number of steps, so it should quickly be able to catch up to the needed bond dimension.

tenpy/algorithms/tebd.py
```diff
@@ -591,6 +591,264 @@ def __init__(self, psi, model, options):
...
+        self.psi.set_B(i0, A_L, form='A')
+        self.psi.set_SL(i1, S)
+        self.psi.set_B(i1, B_R, form='B')
+
+        return trunc_err
+
+
+def _eig_based_svd(A, need_U: bool = True, need_Vd: bool = True, inner_labels=[None, None],
+                   trunc_params=None):
+    """Computes the singular value decomposition of a matrix A, but via diagonalization of
```
jhauschild, line 786: See above on one-line summary.

tenpy/algorithms/tebd.py
```diff
@@ -591,6 +591,264 @@ def __init__(self, psi, model, options):
...
+            # need to be more elaborate in this case ; the new leg on U as computed below
+            # would not be contractible with the new leg in Vd ...
+            raise NotImplementedError from None
+        
+        U, S, _, _, _ = _eig_based_svd(
+            A, need_U=True, need_Vd=False, inner_labels=inner_labels, trunc_params=trunc_params
+        )
+        _, S, Vd, trunc_err, renormalize = _eig_based_svd(
+            A, need_U=False, need_Vd=True, inner_labels=inner_labels, trunc_params=trunc_params
+        )
```
jhauschild, line 810: I don't think that's completely correct, we need to get Vd from U and S.
At least phases of U/Vd are arbitrary, such that `M != U S Vd` in most cases.

There might also be a danger that S from the different sides are not in corresponding order in case of degenerate singular values + machine precision noise? But I'm not sure if this is an issue (beyond the phases), since the singular vectors themselves are not uniquely defined in that case, you just need any ONB of that subspace...

tests/test_tebd.py
```diff
@@ -109,3 +127,5 @@ def test_RandomUnitaryEvolution():
     eng.run()
     print(eng.psi.chi)
     assert tuple(eng.psi.chi) == (16, 8)
+
+
```
jhauschild, line 131: add a test for _eig_based_svd() or at least a big, fat warning in the functions doc string  (and better an `raise NotImplementedError` in the function code if we don't fix it for now).

jhauschild (changes requested):
Thanks, looks great! Apart from the `_eig_based_svd` I have only minor doc-string style comments.

Jakob-Unfried:
Will look into the comments soon, thanks.

I did a few quick and dirty performance benchmarks that seem to indicate poor performance, but i will have to test more carefully.
Dont merge until i get back on this.

tenpy/algorithms/tebd.py
Jakob-Unfried: I am afraid about what happens when we start from a low bond dimension MPS, e.g. a bond dimension. If we have a large on-site Hilbertspace, which is exactly where QR TEBD may be advantageous, we have $d^2$ schmidt values in the exact evolved state but we keep only 2 singular values, if `expand <= 1` as it probably will be. I worry that the schmidt values will not have sufficiently decayed in this case, since the horizontal axis of the entanglement spectrum is quite coarse.

In any case, such problems would manifest in large truncation errors, so they would be detected. So I will remove this.

tests/test_tebd.py
Jakob-Unfried: You only encounter this if you set the non-default option `use_eig_based_svd=True` in the `QRBasedTebdEngine`s config. Have documented a small non-fat warning there. Or do you want a fatter warning?

jhauschild: looks good to me now.

tenpy/algorithms/tebd.py
jhauschild: If you are really worried by this, you can increase the `max(1, ...)` to `max(10, ...)` in line 738 of tebd.py; I think the rest should go through. This will help to much faster increase initially tiny when needed, but the downside is that ithe total expanded size (before truncation) could become significantly larger than `expand * old_chi` if you have many tiny blocks. We could also make it a full parameter (rather than a hard-coded magic parameter).

jhauschild:
I added a comment regarding CBE again. Otherwise ready to merge from my side. Thanks for polishing everything up!

tenpy/algorithms/tebd.py
Jakob-Unfried: I will make it a paramter 👍

Issue #220, opened 2023-04-19
Need to refine Y0 projection in QR-based TEBD


Jakob-Unfried:
I will shortly fix this problem myself.
I am creating this issue mainly as documentation of the problem.

When [creating the initial isometry `Y0`](https://github.com/tenpy/tenpy/blob/fb33509f4448813f32f58c02ef521b11e3555075/tenpy/algorithms/tebd.py#L716),
we currently make an arbitrary choice to keep the first vectors within each block

https://github.com/tenpy/tenpy/blob/fb33509f4448813f32f58c02ef521b11e3555075/tenpy/algorithms/tebd.py#L764

Those vectors can all be zero, e.g. if you start the simulation with a product state and are unlucky about how its dynamcis cooperates with the convention for the order of the basis (will push a test that exhibits this behavior).
Then, `Y0` will be all zeros, which will eventually cause an svd of `Xi`, which is all zeros and yield a bunch of `NaN`s when `svd_theta` tries to normalize the singular values.

Jakob-Unfried:
I will fix this by projecting to those slices with the largest norm, instead of the first few slices.

PR #221, opened 2023-04-19
Fix Y0 projection in QR-based TEBD

Jakob-Unfried:
See #220


Jakob-Unfried:
@jhauschild 
Any idea why the build fails?
It says `ModuleNotFoundError: No module named 'numpy'` when building, even before pytest starts...

Jakob-Unfried:
The tests pass for me locally, lets wait and see if the github actions resolve themselves...

PR #223, opened 2023-04-21
Hotfix setup.py for pip 23.1 compatibility

Jakob-Unfried:


Jakob-Unfried:
In addition to the CI tests, building from source is successful on my machine.

PR #232, opened 2023-04-25
Mpi parallel dmrg plus hc

Jakob-Unfried:
@jhauschild you wanted to just delete the branch, right?


jhauschild:
This is included in the excitation_science branch

Issue #213, opened 2023-04-06
MPS norm preserved in imaginary-time TEBD even with `preserve_norm = False`


jianshaokai:
Hello, I am trying to run a simple TEBD calculation on the TF Ising model with imaginary time evolution and without preserving the norm of the state between steps. Here is my code to set up the calculation:

```python
model_params = {
    'J': 1. , 'g': 1.,  # critical
    'L': 20,
    'bc_MPS': 'finite',    
}
M = TFIChain(model_params)
psi = MPS.from_lat_product_state(M.lat, [['up']])
tebd_params = {
    'preserve_norm': False,
    'N_steps': 1,
    'dt': 0.05j,
    'order': 4,
    'trunc_params': {'chi_max': 100, 'svd_min': 1.e-12}
}
eng = tebd.TEBDEngine(psi, M, tebd_params)
```

Note that I used imaginary `dt`. I then run the calculation and check the norm of the final state:
```python
while np.abs(eng.evolved_time) < 10.:
      eng.run()
psi.norm
```

I find that this returns 1.0, which is not what I expect. I also tried using `eng.calc_U` with `type_evo = imag` and `dt = 0.05`, but I get the same result. To troubleshoot, I went to `run_evolution` in line 258 of algorithms/algorithm.py and added `print(self.psi.norm)` both before and after the call to `self.evolve`. I expect that `evolve` should change the norm after each step, but this is apparently not the case, as it prints out 1.0 each time. Do you have any suggestions on how I could troubleshoot this further?

Jakob-Unfried:
Indeed, `update_bond` does not update `self.psi.norm` unlike `update_bond_imag`.
In principle, we can fix that and add a line like `self.psi.norm *= renormalize`.
If we do, we should also do it in #205 .

TEBD with imaginary time step and a finite system should not call `update_bond` (`update_bond_imag` would be better) though, since the non-unitary gates do not preserve the canonical form, right @jhauschild ?

Should we introduce safeguards or warnings so this does not happen?

Jakob-Unfried:
Fixed by #247

PR #247, opened 2023-05-08
Keep track of MPS norm in update_bond()

Jakob-Unfried:
See issue #213


jhauschild:
No, I thing this is fine when you added it to the change log.

PR #259, opened 2023-05-24
Npc singledispatch

Jakob-Unfried:


Jakob-Unfried:
Using singledispatch does not work after all...
We can not use the class that is corrently defined for the annotation

PR #263, opened 2023-06-07
fix #262 that memory usage had wrong units

### LINKED ISSUE #262: Memory Usage on Mac
Tenpy reports 1024 times peak memory usage on Mac than correct because in:

https://github.com/tenpy/tenpy/blob/main/tenpy/tools/process.py#L42

the line
`resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
`

returns in bytes rather than kilobytes on Mac, e.g. see https://stackoverflow.com/questions/938733/total-memory-used-by-python-process/7669482#7669482

jhauschild:
See #262

### DIFF
--- a/tenpy/tools/process.py
+++ b/tenpy/tools/process.py
@@ -18,6 +18,7 @@ you might also try the equivalent :func:`mkl_get_nthreads` and :func:`mkl_set_nt
 import warnings
 import ctypes
 from ctypes.util import find_library
+import sys
 
 __all__ = [
     'memory_usage', 'load_omp_library', 'omp_get_nthreads', 'omp_set_nthreads', 'mkl_get_nthreads',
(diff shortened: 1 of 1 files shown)


Jakob-Unfried:
The failing doctest should be fixed by #264.

Jakob-Unfried:
On my Macbook with Intel chip, I see the bug in the old version (current main).
For `examples/d_dmrg.py` the first DMRG run reports `69352.0MB`.

After the commit, the reported memory usage agrees with what i can see in `htop`.
It reports  around `67MB` for the finite Ising model DMRG run.
On the TUM workstation running Ubuntu, the same simulation reports around `98MB`.

It would be good to have someone with a M1 and/or M2 chip test it as well, but IMO, we can merge this.

PR #267, opened 2023-06-14
Expectation value class

wilhelmkadow:


jhauschild:
Thanks again, @wilhelmkadow for contributing this! 
I've refactored it a bit further and introduced a separate `BaseEnvironment` to allow the MPSEnvironment, but not the MPOEnvironment being derived from the MPSExpectationValues.

PR #273, opened 2023-06-27
Add phi_ext to more models

Jakob-Unfried:
- Expand documentation for models that already support phi_ext
- Add phi_ext to fermions_spinless and hubbard models


Jakob-Unfried:
You are of course right regarding the check in coupling_strength_add_ext_flux.
Without it though, the phi_ext parameter for models on 1D lattices silently ignores phi_ext.

To fix this, I added a different check to coupling_strength_add_ext_flux, that checks if the number of specified fluxes matches the dimension of the lattice.
This is implicitly assumed by the implementation anyway.

Since this is not what you agreed to above, could you confirm the merge again?

PR #278, opened 2023-07-07
Import structure

Jakob-Unfried:
Reduce namespace structure to two "public" API levels and the "private" level:
1. The top-level namespace `tenpy` e.g. `tenpy.MPS` or `tenpy.tensordot`. This involves choices, which functions/classes to expose.
2. The subpackages, e.g. `tenpy.algorithms.DensityMatrixMixer`. All "public" (i.e. not starting with an underscore) object are exposed at the level of their sub-package.
3. The concrete module, e.g. `tenpy.linalg.charges._partial_qtotal` to access all, including "private" objects

Unless I am very much mistaken, these changes do not break backwards-compatibility, e.g. explicitly import objects from their module like `from tenpy.networks.mps import MPS` will still work and the shorter `from tenpy import MPS` is only an added alternative

### The following questions arise:
- [x] Is the proposed structure above sensible?
- [x] Are the choices for which objects get to be exposed in `tenpy.__init__` sensible (see commits)?
      While this PR is a draft, I provided the "negative" list of objects *not* chosen
      also in `tenpy.__init__`.
- [x] should `tenpy.networks.site.kron` be in `tenpy.linalg.np_conserved`?
- [x] should `tenpy.networks.site` be in `tenpy.models`?
- [x] for `tenpy.linalg.lanczos.lanczos` there is a name conflict which clashes with the above philosophy,
      i.e. does `from tenpy.linalg import lanczos` import the module or the function?
      should rename one of the two.
      Suggestion: rename the module to `tenpy.linalg.krylov_based`.
      I already made this change, so the tests can run, but we can decide to solve it some other way.
- [x] how to adjust the module-level docstrings of the `__init__.py` to make sphinx show this stuff

### Remaining todos after settling questions above:
- [x] Populate the `__all__` lists in top-level and submodule `__init__.py` files.
      I dont want to do this twice, so i did not do it yet.
- [x] Resolve name conflict of  (module vs function)


Jakob-Unfried:
Actually, renaming the lanczos module is breaking backwards compatibility....
So we should separately decide what we want to achieve for v1.0 and how to transition

PR #288, opened 2023-08-21
Generalize Mixers

Jakob-Unfried:
The reason for this PR is to allow the controlled bond expansion to be implemented as a `Mixer`.

# Main changes
- Implement mixers in `tenpy.algorithms.mps_common`
  -> Offer backwards-compatible wrappers in `tenpy.algorithms.dmrg`
- Change methods of the Mixer interface: two different methods (`mixed_svd_2site` and `mix_and_decompose_1site`) for decomposing single-site or two-site theta to replace `peturb_svd`.
  Since the current `perturb_svd` has different signatures for those cases, any users would need
  to distinguish these cases anyway.
  Removed the `update_sites` class attribute.
- Move the implementation of using a "1-site mixer" to decompose a 2-site wavefunction from the
  TwoSiteDMRGEngine to the Mixer base class, as the `mix_and_decompose_2site` method.
- Change docstring summary to cover a broader view that is not specialized on application in DMRG
- Change docstrings to emphasize that the physical state, the wavefunction and the singular values
  are unchanged.
  E.g. reformulate "amplitude; Singular values are perturbed on that order of magnitude.".
- Put longer explanations of how the mixers work in the Notes section of the docstring
- Anticipate mixers where (i) "amplitude" makes no sense (always same strength) or
  (ii) that do not decay / are never turned off
  -> class variables for defaults of `amplitude`, `decay` and `disable_after`
  -> allow `None` for those attrs

# Follow-up TODO list
- [x] Discuss differences between SubspaceExpansion and DensityMatrixMixer in docs. Advantages? Costs?
  -> Make SubspaceExpansion the default and sell it as such?
- [x] Do we need an interface that maps e.g. `--A--Th--` to `--A==Th--` or similar?
      I.e. do mixing without doing an SVD?
- [x] Adjust usage in the notebooks, e.g. in 12_dmrg_mixer.ipynb
- [x] Mixer seems to handle qtotal differently than the pure SVD alternatives in `XxxSiteDMRGEngine.mixed_svd`
- [x] Warn on single-site DMRG with a "two-site" mixer?
- [x] Make `qtotal_LR` work the same as with `svd_theta`, i.e. auto-detect on a single `None` entry.
- [x] Should `S_inv_cutoff` be decreased for all mixers? Not needed if we only do `mix_and_decompose_1site` right, e.g. in single-site DMRG (or future single-site TDVP)?


jhauschild (changes requested):
I looked at the changes. A good first step!

Let me add a few more points that I think should be done in the refactoring:
- [x] Move the `self.mixer` attribute, `mixer_activate` and `mixer_cleanup` into the Sweep class (including the mixer parameters....)
- [x] Also the `chi_list_reenables_mixer`.
- [x] Allow the initializing engine to define other default mixer parameters?
       Currently, they're mostly chose for finite DMRG to decay from initial 1.e-5 to 1.e-10 in 15 sweeps. For inifinte DMRG, it might actually help to keep the mixer on longer?
       TDVP should probably set `decay = disable_after=None`, while those parameters makes sense for the variational algorithms (DMRG, Compression/ApplyMPO).
- [x] Rewrite `mixer_cleanup` to do the "proper" thing rather than the hack to perform another SVD sweep: do SVDs of the psi._S and absorb them into the neighboring tensors.
A bit tricky how to update the environments, though. We could
    1. properly update all of them, even though in most cases that's probably not needed?
    2. just update the environments on the very left/right saved by get_resume_data() and clear the rest to be recomputed on demand.
  
  I think I'd go with the second option here.


Further, we need to check and test whether things actually work for 
- [x] the VariationalCompression
- [ ] the VariationalApplyMPO
- [ ] SingleSiteTDVP
- [ ] TwoSiteTDVP

Jakob-Unfried:
I made a design choice in the commit that I am about to push that is up for debate.
We now have the `'mixer'` option in the Sweep class.
We need to encode two different defaults for every Sweep subclass;
- The default for the option itself, i.e. what happens if the key is not set at all
- The meaning of `mixer=True`, i.e. which mixer to use if the user says "Use a mixer, I dont care which"

So far, this is done by an instancecheck in `DMRGEngine.mixer_activate`, which does not generalize well.

I have added a second class attribute `use_mixer_by_default` in addition to `DefaultMixer`, such that we can control them both separately. In particular, to recover the default behavior of the live main branch we need

```python
class TwoSiteDMRGEngine:
    DefaultMixer = mps_common.DensityMatrixMixer
    use_mixer_by_default = False


class SingleSiteDMRGEngine:
    DefaultMixer = mps_common.SubspaceExpansion
    use_mixer_by_default = True
```

Jakob-Unfried:
> 3. We currently set `S_inv_cutoff` to a much higher threshold (`1.e-8` instead of `1.e-15`) when the mixer is activated. I dont think this is needed for the mixers that use true (i.e. 1D) singular values, such as `SubspaceExpansion` or the soon-to-come controlled bond expansion, if used via `mix_and_decompose_1site`. correct?
>    Ideas how we best implement this distinction in `Sweep.mixer_activate`? It depends both on the type of mixer _and_ how the engine calls it.

What about something like this;
```python
class Sweep:
  ...
  @property
  def S_inv_cutoff(self):
    return 1.e-8 if any(isinstance(S, npc.Array) for S in self._S) else 1.e-15
```
It does the same as the current intention (higher cutoff if there are 2D arrays `S`s due to the mixer)  and subclasses may still override.

Jakob-Unfried:
Regarding the computation costs i get the following, assuming `d < w < chi`:

SubspaceExpansion (single-site theta)
- contraction `d^2 w chi^3`
- SVD of `(d chi, w chi)` costs `d w^2 chi^3` [dominant]
- total `d w^2 chi^3`

SubspaceExpansion (two-site theta)
- contraction `d^3 w chi^3`
- SVD of `(d chi, d w chi)` costs `d^3 w^2 chi^3` [dominant]
- total `d^3 w^2 chi^3`

DensityMatrixMixer
- contraction `d^3 w chi^3` [dominant]
- EIG of `(d chi, d chi)` costs `d^3 chi^3`
- total `d^3 w chi^3`

So it seems that for two-site wavefunctions, the DensityMatrixMixer should be cheaper by a factor of the MPO bond dimension.

Jakob-Unfried:
Everything we discussed, I have now either committed here or opened separate issues for (keep branches short-lived).
See the mentions above.

jhauschild (approved):

PR #292, opened 2023-08-29
Enhance iDMRG stability by avoiding inverses of singular values

jhauschild:
While the finite DMRG sweeps moves the orthogonality center and thus should never need inverses of singular values,
the insertion of the iDMRG unit cell *does* require it so far.
Reasoning: iDMRG looks at a segment of the unit cell and has the MPS in center canonical form with A on the left and B on the right. But to insert a unit cell, we need a unit cell of only B tensors on the right, or all A on the left, respectively.
Hence, we apparently need to take inverse of singular values when inserting the unit cells, at least for the initial guess of the local theta: e.g. for the two-site DMRG with a two-site unit cell, we perform updates
```
                                A[0] S B[1]
-> insert unit cell on right    A[0] S B[1] inv(S) A[0] S B[1]
-> update(1,2)                         A[1]    S   B[0]
-> insert unit cell on left     A[0]   A[1]    S   B[0] inv(S) A[1] S B[0]
-> update(0,1)                                     A[0]    S   B[0]
```
Note that we (should) only absorb B tensors in the right environments (and A into the left one, respectively).

The idea of this pull request is to avoid the inverse singular values all together, at least during the iDMRG sweeps.
This happens at the cost of an additional (single-site) SVD shifting the orthogonality center:
E.g. for the `inv(S) A[0] S` to get `B[0]` for the first update, we can actually do an SVD of ` U s V = svd(A[0] S)` with the physical leg bent to the right, and calculate the two-site theta as `S B[1] U V`. Essentially, the `U V` is the unitary part of a polar decomposition, similar in spirit as in VUMPS. Note that the `UV` is an isometry as expected from B by construction as long as `U` is a unitary, which is the case if the dimensions of charge blocks of the `vL` of `A[0] S` are smaller than in the pipe `(p.vR)` going to the right. 
The `s` should just be the singular values in the center, if the state is translation invariant.  If the `S` in the center (for which we try to avoid the inverse) is not diagonal, e.g. for a mixer, we need to incorporate another basis transformation, since the `B[1]` has the `vR` leg *right* of the center S, while the `A[0]` leg has the `vL` leg *left* of the center S. But we  can get that basis transformation again by doing an SVD of the (matrix) `S`, `X Y Z = svd(S)` and get the `B[0]` as  `dagger(X Z) U V` in this case.

This PR implments the suggestion above in the `MPS._scale_B_inverse_free()` method, to be used by `MPS.get_B(i,form='B')` if the saved tensor is in `'A'` form (and similar with A<->B).

### DIFF
--- a/tenpy/networks/mps.py
+++ b/tenpy/networks/mps.py
@@ -3547,7 +3547,7 @@ class MPS(BaseMPSExpectationValue):
         M = self.get_B(L - 1, form)
         M = npc.tensordot(R, M, axes=['vR', 'vL'])
         if self.bc == 'segment':
-            # also neet to calculate new singular values on the very right
+            # also need to calculate new singular values on the very right
             U, S, VR_segment = npc.svd(M.combine_legs(['vL'] + self._p_label),
                                        cutoff=cutoff,
                                        inner_labels=['vR', 'vL'])
@@ -5442,17 +5442,17 @@ class MPSEnvironment(BaseEnvironment, BaseMPSExpectationValue):
         return contr * self.bra.norm * self.ket.norm
 
     def _contract_LP(self, i, LP):
-        LP = npc.tensordot(LP, self.ket.get_B(i, form='A'), axes=('vR', 'vL'))
+        LP = npc.tensordot(LP, self.ket.get_B(i, form='A', avoid_S_inverse=True), axes=('vR', 'vL'))
         axes = (self.ket._get_p_label('*') + ['vL*'], self.ket._p_label + ['vR*'])
         # for a ususal MPS, axes = (['p*', 'vL*'], ['p', 'vR*'])
-        LP = npc.tensordot(self.bra.get_B(i, form='A').conj(), LP, axes=axes)
+        LP = npc.tensordot(self.bra.get_B(i, form='A', avoid_S_inverse=True).conj(), LP, axes=axes)
         return LP  # labels 'vR*', 'vR'
 
     def _contract_RP(self, i, RP):
-        RP = npc.tensordot(self.ket.get_B(i, form='B'), RP, axes=('vR', 'vL'))
+        RP = npc.tensordot(self.ket.get_B(i, form='B', avoid_S_inverse=True), RP, axes=('vR', 'vL'))
         axes = (self.ket._p_label + ['vL*'], self.ket._get_p_label('*') + ['vR*'])
         # for a ususal MPS, axes = (['p', 'vL*'], ['p*', 'vR*'])
-        RP = npc.tensordot(RP, self.bra.get_B(i, form='B').conj(), axes=axes)
+        RP = npc.tensordot(RP, self.bra.get_B(i, form='B', avoid_S_inverse=True).conj(), axes=axes)
         return RP  # labels 'vL', 'vL*'
 
     # methods for Expectation values
--- a/tenpy/networks/mps.py
+++ b/tenpy/networks/mps.py
@@ -1843,8 +1843,9 @@ class MPS(BaseMPSExpectationValue):
                       down='down',
                       lonely=[],
                       lonely_state='up',
-                      bc='finite'):
-        """Create an MPS of entangled singlets.
+                      bc='finite',
+                      form='B'):
+        """Create an MPS of entangled singlets, also known as valence bold solid.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


jhauschild:
Open ToDo's:
- [ ] The uncommented `warnings.warn("inverse singular values!")` in `MPS.get_B()` still apear at a lot of different places. Go through them and check whether we can avoid them.
- [ ] I realized that the iDMRG sweep ends with the update on sites (1, 2), not (0, 1). This implies that one needs the inverse S in the first update of the following sweep, or when calculating expectation values on sites (0, 1).
Should we update the iDMRG sweep schedule to start on sites (1, 2) and end with sites (0, 1)?
- [ ] I checked that all the tests still run, so at least iDMRG still converges. But before merging this PR, we should carefully check wether it actually helps in the sense of reducing the number of DMRG sweeps needed, especially for critical systems.

tenpy/networks/mps.py
```diff
@@ -2026,14 +2035,68 @@ def get_B(self, i, form='B', copy=False, cutoff=1.e-16, label_p=None):
...
         return B
 
+    def _scale_B_inverse_free(self, i, A, A_to_B=True):
+        """Change from A to B canonical form with extra SVD instead of inverses of singular values.
+
+        If `A_to_B` = True, we transfrom the given `A` to `B`-form.
+        If `A_to_B` = False, we assume we're given `B` form and return `A` form.
+
+        The returned B is an isometry, say for A_to_B (analogously statements for B_to_A),
+        if 1) the
```
Jakob-Unfried, line 2071: docs incomplete?

tenpy/networks/mps.py
```diff
@@ -2026,14 +2035,68 @@ def get_B(self, i, form='B', copy=False, cutoff=1.e-16, label_p=None):
...
+                # mixer is on, so SL is a 2D array.
+                # The `A` includes the SL, so has the leg *left* of SL.
+                # `B` should have leg *right* of SL, so we need to apply the basis trafo from
+                # left of SL to right of SL on the `vL` leg of `th`
+                U, _, Vd = npc.svd(SL.transpose(['vL', 'vR']), inner_labels=['vR', 'vL'])
+                V_Ud = npc.tensordot(U, Vd, axes=['vR', 'vL']).conj().itranspose(['vR*', 'vL*'])
+                th = npc.tensordot(V_Ud, th, axes=['vL*', 'vL']).ireplace_label('vR*', 'vL')
+            th = th.combine_legs([['vL'], self._p_label + ['vR']], qconj=[+1, -1])
+        else:  # convert B to A
+            #  assert False, "where do we need this????"
```
Jakob-Unfried, line 2088: remove?

tenpy/networks/mps.py
```diff
@@ -2147,38 +2213,158 @@ def get_theta(self, i, n=2, cutoff=1.e-16, formL=1., formR=1.):
             if self.form[j % self.L] is None:
                 raise ValueError("can't calculate theta for non-canonical form")
         if n == 1:
-            return self.get_B(i, (1., 1.), True, cutoff, '0')
+            return self.get_B(i, (1., 1.), True, cutoff, '0', avoid_S_inverse)
```
Jakob-Unfried, line 2216: Use `(form_L, form_R)`

Jakob-Unfried (commented):
Looks good to me!

I double checked that this gives the correct A/B without mixers.
I am "only" 95% convinced that it is ok to do this with the mixer.

Jakob-Unfried:
My explanation for why this is correct.
Maybe we want to put this in the docs under notes, or as comments in the code or something?

```
With a mixer, the SL is not a diagonal matrix.
We use an SVD of it to derive what to do::

    |   --SL--   =   --X--Y--Z--

with isometric ``X, Z`` and diagonal ``Y``.
We compute an SVD of the following matrix::

    |   --M--   :=   --hc(Z)--hc(X)--theta--   =   --U--S--V--
    |     |                            |                   |

But we can obtain a different SVD from the canonical form, using ``theta = SL @ B``::

    |   --M--   =   --hc(Z)--Y--Z--B--
    |     |                        |

where ``Z @ B`` is the right isometry.
Two SVDs of the same matrix are related by a gauge transformation, a diagonal matrix ``phi`` of complex phases such that::

    |   --U--   =   --hc(Z)--phi--     ;     S = Y     ;     --V--   =   --hc(phi)--Z--B--
    |                                                          |                       |

Therefore we have::

    |   --U--V--   =   --hc(Z)--phi--hc(phi)--Z--B--   =   --B--
    |        |                                   |           |
    
```

PR #235, opened 2023-04-25
Spectral function simulation class

Jakob-Unfried:
Opening this PR so we can discuss and see it in the v1.0 milestone.

Issue #282, opened 2023-07-18
missing jordan-Wigner strings for some functions


jhauschild:
Some functions do not yet properly take into account jordan-Wigner strings - or at least raise Errors, if there should be some.

- [ ] `MPS.apply_local_op('C')` runs in current main branch without error. There's already a fix to actually include JW strings in the `excitations_science_and_vumps` branch, but that's still not yet merged into main.
- [ ] MPS.expectation_value() currently also doesn't check wether it get's a `C` operator to raise an error.
- [ ] On the other hand, MPSEnvironment.expectation_value('C') could in principle be legit when the state have different charges...

PR #293, opened 2023-08-29
Spectral functions draft

jhauschild:
Opening a draft pull request for the work of @P-thoma
Since this is an enhanced continuation of #235, I'll close that other PR, we can discuss things here instead.


tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+            if self.model.lat.site(i).op_needs_JW(op):
+                for j in range(i):
+                    self.psi.apply_local_op(j, 'JW')
+            self.psi.apply_local_op(i, op)  # TODO: check if renormalize=True makes sense here
+        else:
+            ops, i_min, _ = self.psi._term_to_ops_list(operator_t0, True)  # applies JW string automatically
+            for i, op in enumerate(ops):
+                self.psi.apply_local_op(i_min + i, op)
+
+    def prepare_results_for_save(self):
```
jhauschild, line 276: Nice function! You can move this code directly into the base `Simulation` class.

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+            spectral_function_t = np.array(spectral_function_t)
+
+        # multiply evolution of bra (eigenstate) into spectral function
+        phase = np.exp(1j * self.gs_energy * self.engine.evolved_time)
+        spectral_function_t = spectral_function_t * phase
+
+        return spectral_function_t
+
+
+class SpectralSimulationExperimental(SpectralSimulation):
```
jhauschild, line 351: When you tested this functionality, you can merge it into the main class :wink:

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+        """Calculate the overlap :math:`<psi_0| e^{iHt} op2^j e^{-iHt} op1_idx |psi_0>` between
+        op1 at MPS position idx and op2 at the MPS position j"""
+        self.logger.info("calling m_spectral_function")
+        operator_t = self.operator_t
+        env = self.get_mps_environment()  # custom method for subclass Experimental
+        # TODO: get better naming convention, store this in dict ?
+        if isinstance(operator_t, list):
+            for i, op in enumerate(operator_t):
+                if isinstance(op, str):
+                    results[f'spectral_function_t_{op}'] = self._m_spectral_function_op(env, op)
```
jhauschild, line 308: Yes, we need better naming conventions here. 
In general, I think we should allow more generic operator terms to be measured (as we discussed with Frank as well), we want generic `<psi | A(t) B|psi>`, with A/B sometimes being multi-site operators, see also the comment on `apply_operator_t0_to_psi` above.
I think a convenient parameter conventione might be: either `operator_t` is a single dict specifying the operator / term A (with MPS /lattice index and op name), or you have a dict `{opname_1: {...}, opname_2: {...}}`, which each of the sub dictionaries as you would have for a single operator.
Of course, the same convention for specifying a single term/local operator should be used for the both the t0 and t operators.

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+        else:
+            spectral_function_t = list()
+            for i in range(self.psi.L):
+                term_list, i0, _ = env._term_to_ops_list([('Id', 0), (op, i)], True)
+                # this generates a list from left to right
+                # ["JW", "JW", ... "JW", "op (at idx)"], the problem is, that _term_to_ops_list does not generate
+                # a JW string for one operator, therefore insert Id at idx 0.
+                assert i0 == 0  # make sure to really start on the left site
+                spectral_function_t.append(env.expectation_value_multi_sites(term_list, i0))
+                # TODO: change when :meth:`expectation_value` of :class:`MPSEnvironment` automatically handles JW-string
```
jhauschild, line 341: Yes, again, we should implement #282 and adjust code here. (saying this again here to get another cross-reference in the github issue).

jhauschild (commented):
Looks pretty good! And quite some code that you wrote there, great work!

Please don't be discouraged from the plenty of comments, it's just a lot of minor details where I wanted to point out a few things that I would have done differently, or coding conventions/guidelines in TeNPy.
If you have any questions, I'm of course happy to discuss everything in person :wink:

A few general comments:
- [ ] I'm not sure how detailed you read the [TeNPy coding guidelines](https://tenpy.readthedocs.io/en/latest/contr/guidelines.html). When I added comments on formatting (e.g. for the newlines in docstrings) that you don't find there, please feel free to expand those guidelines.
- [ ] We need to add some tests, ideally fine-grained for small parts rather than just one test running a full SpectralFunctionSimulation. For example you can easily add a test for reciprocal basis that it works correctly on the predefined lattices. Doing so might uncover further issues, for example the next point in this list.
Further, you should add tests for the linear prediction e.g. checking that sums of exponentials and cosines/sines get predicted correctly. I know that you already ran some checks, so all you have to do is formalize those into the test setup. Pytest has quite good docs, e.g. on [how to get started](https://docs.pytest.org/en/7.3.x/getting-started.html#get-started). 
- [ ] The Ladder is a 1D lattice (with a single basis vector and `Ladder.dim == 1`), but the `Ladder.unit_cell[i]` and `Ladder.basis[0]` are 2D vectors. This is mostly a hack to allow proper plotting in 2D, but I think it screws with your way of calculating the reciprocal basis, and needs a separate handling. 
- [ ] For fast tests of the simulation classes, we can use a "dummy" simulation which doesn't actually do the evolution with say TEBD, but just updates the evolved time etc. You can take a look into `test_simulation.py` for inspiration.
- [ ] This is probably more my fault for still not merging #266, but just to document/comment on this here: there is some significant overlap with the `OrthogonalExcitations` code there also loading ground states from file etc. Ideally, we should try to avoid duplicated code to allow simpler maintenance of the code. The same is true when we want to think about segment boundary conditions (you have a TODO for that...).

You wrote so much that it took me several hours to go through it, and I didn't get to the `post_processiong.py`. I'll take a look later, hopefully tomorrow. Impressive work from your side!

examples/yaml/minimal_SpectralSimulation.yml
```diff
@@ -0,0 +1,35 @@
...
+algorithm_params:
+    trunc_params:
+        chi_max: 120
+    dt : 0.1
+    N_steps : 2  # measurements every dt*N_steps
+
+post_processing_params:
+    output_filename: post_processed_results.h5
+    linear_prediction:
+        m: 50
```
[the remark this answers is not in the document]
p-thoma: Changed this to relative parameters (now under `spectral_function_params`):

- `rel_prediction_time`  
- `rel_num_points`

tenpy/networks/mps.py
```diff
@@ -5289,6 +5306,26 @@ def _contract_with_RP(self, C, i):
         return C
 
 
+class MPSEnvironmentJW(MPSEnvironment):
+    """Class similar to :class:`MPSEnvironment`, but overwriting the method
+    :meth:`_contract_LP` in a way that the JW string is automatically added when
+    contracting the left environment only -> expectation values should only
+    be calculated from left to right with this class. This works for finite MPS."""
```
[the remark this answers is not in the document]
p-thoma: Done, however, I noticed that `MPSEnvironment` does not have an attribute `bc` -- instead there's `ket.bc` and `bra.bc`. I'm not sure if there's a special case, where `bra.bc`  !=  `ket.bc`, otherwise does it make sense to set a `bc` attribute for the `MPSEnvironment` (or respectively setting this in `BaseEnvironment`)  ?

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+                else:
+                    self.psi_ground_state = gs_data['psi']
+                delattr(self, 'gs_data')  # possibly free memory
+            else:
+                self.logger.warning("No ground state data is supplied, calling the initial state builder on\
+                                     SpectralSimulation class. You probably want to supply a ground state")
+
+                super().init_state()  # this sets self.psi from init_state_builder (should be avoided)
+                self.psi_ground_state = self.psi.copy()
+                delattr(self, 'psi')  # free memory
```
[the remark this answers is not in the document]
p-thoma: This will definitely need a rework (as the logic is over-complicated). 
However, the point is that if `psi` is set manually that should translate to setting the ground-state.
And psi needs to be removed (intermediately) in order to be compatible with checkpointing (otherwise we might apply the initial operator to a time evolved checkpoint again). Alternatively I could overwrite `resume_run` in a way that `init_state` is not called. But maybe it's better to fix the structure in a later version?

p-thoma:
# Update 
As the [code review](https://github.com/tenpy/tenpy/pull/293#pullrequestreview-1717893435) by @jhauschild  was quite a while ago and some things have changed considerably, I want to give an update to keep track of everything. The following is implemented/work in progress:

- [x] Simple version of a `SpectralSimulation` class which allows one to obtain the Spectral Function of a *finite* system in 1D and 2D. Support for automatically loading data/the state to be time evolved from another `Simulation` (e.g. a `GroundStateSearch`). The Spectral Function can be calculated for *onsite* operators although they have to be specified by name (e.g. `'Sz'` for a `SpinHalfSite`).
- [x] Extending the `Lattice` class by a simple class for the Brillouin Zone `BZ` including functions to plot the Brillouin Zone, the reciprocal basis vectors and to reduce points outside the Brillouin Zone back into it
- [ ] A class `DataLoader` to better handle results from a finished simulation (e.g. giving access to the underlying model & lattice and more easily viewing/loading parameters and results
- [ ] post-processing of measurements at the end of a `Simulation`. This is done by wrapping/writing the required functions with/as a post-processing function (i.e. `pp_custom_processing`) which take a `DataLoader` instance as only argument and all additional arguments as kwargs. This is similar in design to custom measurement functions (hence post-processing functions can also be added to the `.yaml` file)
- [ ] Example Notebooks (I already have plenty at least for the Spectral Simulation)

## General TODO/Improvements

- [ ] writing extensive tests ([comment](https://github.com/tenpy/tenpy/pull/293#pullrequestreview-1717893435))
- [ ]  remove workaround because `apply_local_op()` ignores the JW string ([comment](https://github.com/tenpy/tenpy/pull/293#discussion_r1385527817) and [comment](https://github.com/tenpy/tenpy/pull/293#discussion_r1385613927)), related to #282
- [ ] allow to apply custom operators more complicated operators ([see](https://github.com/tenpy/tenpy/pull/293#discussion_r1385611881)) and possibly restructure `TimeDependentCorrelation` ([see](https://github.com/tenpy/tenpy/pull/293#discussion_r1463352086)) to allow different states from the ground-state
- [ ] error checks, see also #314
- [ ] as discussed last week with @jhauschild implement the time-evolution using *segment* boundary-conditions (although this is not directly linked to the `SpectralSimulation`
- [ ] sharing code/data loading with #266 

## Current Structure
The majority of the work is done in the following modules

- `simulations.time_evolution`: `TimeDependentCorrelation` only needs testing (although there is still an experimental version using `MPSEnvironmentJW`. `SpectralSimulation` depends highly on  `post_processing`, `spectral_function_tools` and `prediction`
- `simulations.post_processing`: `DataLoader` class will probably need some more development
- `tools.spectral_function_tools`: provides main function to calculate the Spectral Function from a `TimeDependentCorrelation`. Only dependent on `linear_prediction` and `Lattice`
- `tools.prediction`: implementing linear prediction, independent of everything else, but still needs testing
- `models.lattice`:  holding the Brillouin Zone class -> completely independent of other modules, only some tests are missing

### Suggested Merge Plan

Therefore we could try to merge `models.lattice`, `tools.prediction` and the Simulation: `TimeDependentCorrelation` soon, so that the basis is merged and later on, we'd first have to merge the `DataLoader` class from `post_processing` in order to be able to merge the `SpectralSimulation`  class and the `spectral_function_tools` as a final step ?

tenpy/simulations/time_evolution.py
p-thoma: Changed interface to include a TimeDependentCorrelationEvolveBraKet
for having two engines

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +80,347 @@ def final_measurements(self):
...
+    def apply_operator_t0_to_psi(self):
+        # TODO: think about segment boundary conditions
+        # TODO: make JW string consistent, watch for changes in apply_local_op to have autoJW
+        self.operator_t0 = self._get_operator_t0()
+        operator_t0 = self.operator_t0
+        if len(operator_t0) == 1:
+            op, i = operator_t0[0]
+            if self.model.lat.site(i).op_needs_JW(op):
+                for j in range(i):
+                    self.psi.apply_local_op(j, 'JW')
```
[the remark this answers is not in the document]
jhauschild: we need to double-check this after merging the vumps and current main

tenpy/tools/spectral_function_tools.py
```diff
@@ -0,0 +1,258 @@
...
+    lat :
+        a (TeNPy) lattice to plot the correlations on
+    correlations : array-like
+        an array of correlations (in mps_form)
+    pairs: str
+    scale: float
+    color_pos: str
+    color_neg: str
+    color: str
+    zorder: float
```
jhauschild, line 213: docs. 

It might also be nice to have an alternative version that has fixed line width, but changes color according to a colormap (e.g. red-blue if we have positive/negative values). That would probably be a different function, though.

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +82,411 @@ def final_measurements(self):
...
+            self.engine.checkpoint.emit(self.engine)  # set up in init_algorithm of Simulation class
+            # and connects to self.save_at_checkpoint which ultimately calls self.save_results()
+            # if there are no results self.prepare_results_for_save() which calls get_resume_data() which
+            # calls engine.get_resume_data
+
+    def m_correlation_function(self, results, psi, model, simulation, **kwargs):
+        """Equivalent to :meth:`TimeDependentCorrelation.m_correlation_function`."""
+        self.logger.info("calling m_correlation_function")
+        operator_t = to_iterable(self.operator_t)
+        psi_bra = self.engine_bra.psi.copy()  # make copy since algorithm might use grouped bra
```
jhauschild, line 414: This is a small optimization, but you can move the copy to the if self.grouped > 1: case which is the only one where youneed it, isn't it?

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +82,411 @@ def final_measurements(self):
...
+
+    def get_resume_data(self):
+        resume_data = super().get_resume_data()
+        resume_data['psi_ground_state'] = self.psi_ground_state
+        resume_data['gs_energy'] = self.gs_energy
+        return resume_data
+
+    def _connect_measurements(self):
+        """Connect :func:`m_correlation_function` to measurements."""
+        self._connect_measurements_fct('simulation_method', 'm_correlation_function', priority=1)
```
jhauschild, line 175: why do we do it like this, and not add 
`('simulation_method', 'm_correlation_function')` to the class _default_measurements instead?
See the update of the defaul_measurements in RealTimeEvolution as an example

tenpy/simulations/time_evolution.py
```diff
@@ -74,3 +82,411 @@ def final_measurements(self):
         We already performed a set of measurements after the evolution in :meth:`run_algorithm`.
         """
         pass
+
+
+class TimeDependentCorrelation(RealTimeEvolution):
+    r"""A subclass of :class:`RealTimeEvolution` to specifically calculate the
```
jhauschild, line 88: single line. Maybe:
```
Specialized :class:`RealTimeEvolution` to calculate a time dependent correlation function of a ground state.
```
Mention that psi_0 needs to be an eigenstate of H for this variant in the doc string below, and refer to the `TimeDependentCorrelationEvolveBraKet` as generalization for arbitrary initial states

jhauschild (commented):
I have a couple more suggestions that shouldn't take long to fix.

tenpy/simulations/time_evolution.py
p-thoma: In case `Simulation.use_default_measurements` is set to `False` the correlation function would still 
be measured. Otherwise when disabling the default measurements the class  `TimeDependentCorrelation` would 
still work but w/o any useful output. I'll add this to the docstring.

PR #294, opened 2023-09-01
Support move_right=None

Jakob-Unfried:
It is currently not well-defined what the value of the `Sweep.move_right` attribute should be, if the sweep is not actually moving, i.e. if the next update is at the same position.
This happens e.g. in TDVP, where after the last update of a sweep at `i0=0`, the same site `i0=0` is updated again at the start of the next sweep.

Both TDVP engines set ``(i0, move_right) == (0, False)`` for the last update, which contradicts the definition from the docstring of `Sweep`:
```
move_right : bool
        Only set during sweep.
        Whether the next `i0` of the sweep will be right or left of the current one.
```

This causes problems in `Sweep._cache_optimize`:

> ```python
> ...
> if self.n_optimize == 2:
>     kwargs = {
>         'short_term_LP': [i0, i0 + 1],
>         'short_term_RP': [i0, i0 + 1],
>     }
>     if move_right:
>         kwargs['preload_RP'] = i0 + 2
>     else:
>         kwargs['preload_LP'] = i0 - 1
> ...
> ```

, where `not move_right` is interpreted as "we are moving to the left".

This PR addresses this by introducing support a third option for `move_right`, namely `None`, which means "not moving at all, next i0 is the same as the current".
`Sweep._cache_optimize` could then distinguish three cases.

I have added a "goal" version in the first commit, which breaks backwards compatibility and then introduced backwards compatible wrappers in the second commit which issue an appropriate warning.
We can then undo the changes from the second commit when moving to v1.0.

### DIFF
--- a/doc/changelog/latest.rst
+++ b/doc/changelog/latest.rst
@@ -46,6 +46,10 @@ Changed
   :meth:`~tenpy.models.mixed_xk.MixedXKModel.add_inter_ring_interaction`, and
   :meth:`~tenpy.models.mixed_xk.MixedXKModel.add_intre_ring_interaction` to vary with `x`.
 - Renamed the module :mod:`~tenpy.linalg.lanczos` to :mod:`tenpy.linalg.krylov_based`.
+- The :attr:`~tenpy.algorithms.mps_common.Sweep.move_right` attribute of
+  :class:`~tenpy.algorithms.mps_common.Sweep` now supports a third value, ``None``, in addition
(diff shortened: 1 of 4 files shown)


Jakob-Unfried:
Left to do:
- [ ] Deal with `Sweep._update_env_inds`, I am not sure what should be done here...
  @jhauschild do you know what this should do in case of `move_right is None`?
- [ ] Document the backwards-compatible solution properly
- [ ] changelog

PR #305, opened 2023-09-25
Predict ram

### LINKED ISSUE #48: function to estimate memory requirements
It would be helpfull if simulations/algorithms could estimate roughly how much memory they might need, such that one can reserve a suitable amount when running it on a computing cluster.

jangeiger:
This code predicts the RAM usage required to perform an algorithm in TeNPy (for example, DMRG).
Closes #48

The following features are implemented:
- [x] Implemented usage for MPS, MPO and environments
- [x] Implemented saving due to conservation laws
- [x] Extracted saving factor for particle number conservation in the Bose-Hubbard model

It is compatible with infinite and finite MPS system sizes and was benchmarked for the following cases:

**1) Finite 1D:**
- [x] Tested with finite 1D Ising model (Parity conservation)
- [x] Tested with finite 1D Bose-Hubbard model (Particle number conservation)

**2) Infinite 1D:**
- [x] Tested with infinite 1D Bose-Hubbard model (Particle number conservation)
- [x] Tested with infinite 1D Ising model (Parity conservation)

**3) 2D Lattice:**
- [x] Tested with 2D triangular Heisenberg model

Remark: All RAM tests were performed on the TUM cluster running on Linux. For finite size MPS, the length L was varied; for iMPS, the bond dimension was varied. The ram was extracted using psutil as well as tracemalloc.

### DIFF
--- a/tenpy/models/hubbard.py
+++ b/tenpy/models/hubbard.py
@@ -89,6 +89,30 @@ class BoseHubbardChain(BoseHubbardModel, NearestNeighborModel):
         model_params.setdefault('lattice', "Chain")
         CouplingMPOModel.__init__(self, model_params)
 
+    def estimate_RAM_saving_factor(self):
+        """Returns the expected saving factor for RAM based on charge conservation.
+        For the BoseHubbardChain this factor was found to be between 1/7 and 1/10, therefore it is by default to 1/8 (for particle number conservation).
+
+        Returns
+        -------
+        factor : int
+            saving factor, due to conservation
+
+        Options
+        -------
+        .. cfg:configoptions :: Model
+
+            saving_factor :: None | int
+                Quantizes the RAM saving, due to conservation laws. By default it is 1/8 for the BoseHubbardChain. However, this factor might be overwritten, if a better approximation is known. In this case one can pass it via the argument ''saving_factor'' to the model.
+
+        """
+        chinfo = self.lat.unit_cell[0].leg.chinfo
+        savings = 1
+        for mod in chinfo.mod:
+            if mod == 1:
+                savings *= 1/8 # this is what we found empirically
+        return self.options.get("saving_factor", savings)
+
 
 class FermiHubbardModel(CouplingMPOModel):
     r"""Spin-1/2 Fermi-Hubbard model.
--- a/tenpy/models/hubbard.py
+++ b/tenpy/models/hubbard.py
@@ -98,12 +98,6 @@ class BoseHubbardChain(BoseHubbardModel, NearestNeighborModel):
         factor : int
             saving factor, due to conservation
 
-        Options
-        -------
-        .. cfg:configoptions :: Model
-
-            saving_factor :: None | int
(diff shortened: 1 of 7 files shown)


tenpy/models/hubbard.py
```diff
@@ -89,6 +89,30 @@ def __init__(self, model_params):
...
+        Returns
+        -------
+        factor : int
+            saving factor, due to conservation
+
+        Options
+        -------
+        .. cfg:configoptions :: Model
+
+            saving_factor :: None | int
```
Jakob-Unfried, line 105: Explicitly say that this is only used for RAM estimation via :func:`...Simulation.estimate_RAM`?

tenpy/models/model.py
```diff
@@ -254,9 +254,33 @@ def update_time_parameter(self, new_time):
...
+        for mod in chinfo.mod:
+            if mod == 1:
+                savings *= 1/4 # this is what we found empirically
+            else:
+                savings *= 1/mod
+        return savings
 
 class NearestNeighborModel(Model):
-    r"""Base class for a model of nearest neigbor interactions w.r.t. the MPS index.
+    r"""Base class for a model of nearest neighbour interactions w.r.t. the MPS index.
```
Jakob-Unfried, line 283: `neighbor` is the correct US spelling.

tests/test_predict_ram.py
```diff
@@ -0,0 +1,50 @@
+"""Short test for vmem prediction."""
+# Copyright 2018-2023 TeNPy Developers, GNU GPLv3
```
Jakob-Unfried, line 2: 2023

Jakob-Unfried (commented):
Very nice!
I did not go through the internal logic in detail, it is only heuristics anyway and if it matches in the tests that you conducted, that is good enough for me.

Left some minor comments, let me know if anything is unclear

Jakob-Unfried:
@jangeiger anything missing or can we merge?

jangeiger:
If the code looks good to you, we can merge.

Apart from your comment on the config option, I have implemented everything else. I agree that the saving factor is not a physical parameter. If you wish, we can remove this option. However, for a more accurate prediction, one would probably want to automatically extract it via a small-scale DMRG simulation anyway.

Jakob-Unfried:
While reading this again, I found that `estimate_RAM_saving_factor` looks buggy. Should theta not be an `else` clause that deals with `mod != 1`?

I have two main worries regarding the `saving_factor` as a config option:

- Very few people will know the models and charges intimately enough to give good values. I certainly would not trust myself to give a good guess without running numerical experiments. The whole point of this feature is to move away from peoples intuition and to have some hard-coded heuristics.

- The option will appear in the docs of all models, e.g. [here](https://tenpy.readthedocs.io/en/latest/reference/tenpy.models.toric_code.ToricCode.html#cfg-config-ToricCode). That means that if you give model parameters somewhere, you might encounter this option, and in most cases users should just leave it alone.

I suggest to simply remove the option and rely on the heuristics that you found.

If there is some way to get a better heuristic, we should merge it into tenpy if it concerns one of the models we implement, or the user can simply overwrite `estimate_RAM_saving_factor` for their particular model.

Jakob-Unfried:
Other than that, we are ready to merge =)

jangeiger:
I agree this option is confusing - I have removed it 👍

PR #313, opened 2023-10-26
Fix typos

Jakob-Unfried:


jhauschild (approved):
Wow, that's a lot of typos!
At least, almost all of them have been in the documentation, but a few are even code fixes.
Thanks a lot!

PR #315, opened 2023-11-24
Make splitting of op names consistent in Site

Jakob-Unfried:
We currently use a single space as separator in `get_op` and `get_hc_op_name`, but any whitespace in `op_needs_JW` and `valid_opname`.
We should be consistent.
The docs say "separated by whitespace", so i chose to support any whitespace in all four.
Or should we only do single space @jhauschild ?


jhauschild:
Any whitespace is fine here.

Issue #230, opened 2023-04-25
Set up linting and autformatting


Jakob-Unfried:
We should set up a linter (probably flake8?) that checks moderate code consistency and style requirements
(unused imports, unused variables, line width, ...) as an action on PRs.
Should make it easy for developers to run the linter locally before pushing.

Additionally we should set up an auto-formatter (probably yapf) with stricter rules to run periodically (e.g. monthly)
and reformat the code base.
This could then do whitespace, blank lines, ...

We will need to make the existing codebase comply with those rules, manually.
To avoid merge hell, we will do this after an effort to merge some branches as we move towards v1.0.0

Jakob-Unfried:
Will add flake8 with minimal checks in #249.
Before v1.0, expand the ruleset to cover more, at least line width.

Jakob-Unfried:
Reminder to let the linter check that `tenpy` itself is never imported with absolute imports

Jakob-Unfried:
Reminder to consider moving the checks in `test_consistencies.py` to the linting workflow

Jakob-Unfried:
Reminder to consider newer yapf version

Jakob-Unfried:
Reminder to consider the following disabled part from the pytest github action
```
#        pip install flake8
#        # stop the build if there are Python syntax errors or undefined names
#        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
#        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
#        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```

Jakob-Unfried:
Reminder to update the outdated(?) statements in doc/contr/guidelines.rst

Jakob-Unfried:
Should include more flake8 rules, e.g. F841,F811, the pycodestyle rules (E**), linelength

PR #321, opened 2024-01-19
Update year in copyright notices

Jakob-Unfried:
Two commits:
- Automatically update the copyright year in the docbuild (appears e.g. at the bottom of the readthedocs page)
- Update all copyright notices in the code to 2024 and reactivate the test that checks this (was deactivated in 03b259a)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -8,6 +8,7 @@ import inspect
 import sphinx_rtd_theme
 import io
 import warnings
+from datetime import datetime
 
 # ensure parent folder is in sys.path to allow import of tenpy
 REPO_PREFIX = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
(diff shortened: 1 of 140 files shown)


jhauschild:
Testing the copyright is not a legitimate "test" - the program still runs perfectly fine, it doesn't break just because we have a new year - but the copyright test does.
If anything, we might add a linting rather than a test, that checks on modified/new files that the copyright line is there and "up to date".
Ideally, we should have a github action that auto-formats the code to unify style, and in that process also adds a copyright notice if needed. I guess that should be part of #230

Jakob-Unfried:
That is true for all of `test_consistencies.py` though, right? None of those checks indicate that the code doesnt run anymore?

Either way, I will remove the change to the test and only keep the changed year in this PR

PR #324, opened 2024-01-22
Clean up testing and linting workflows

Jakob-Unfried:


Jakob-Unfried:
@jhauschild as discussed. how do you like the path `tests/linting.py` for the *non*-pytest file?

PR #325, opened 2024-01-22
Fix typos in docs

p-thoma:
A list of typos I (and some credit goes to my IDE) found whilst reading the documentation over the last while :)

### DIFF
--- a/doc/contr/build_doc.rst
+++ b/doc/contr/build_doc.rst
@@ -1,4 +1,4 @@
-Bulding the documentation
+Building the documentation
 =========================
 
 You can use `Sphinx <https://www.sphinx-doc.org>`_ to generate the full documentation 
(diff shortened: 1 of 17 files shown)


p-thoma:
I was also wondering, why the bullet-lists do not get rendered in the contributing section - as this would greatly enhance the readability of this section (I does work for me locally though - so the .rst files containing the lists are correctly formatted)...

Jakob-Unfried:
> I was also wondering, why the bullet-lists do not get rendered in the contributing section - as this would greatly enhance the readability of this section (I does work for me locally though - so the .rst files containing the lists are correctly formatted)...

There is currently something wrong with the readthedocs build, see #322.
I expect that this might just work once we get it to build correctly again...

PR #338, opened 2024-02-15
Clean up docstring of DMRGEngine.run

Jakob-Unfried:


Jakob-Unfried:
@jhauschild any idea why the doc build fails here...?

PR #343, opened 2024-02-20
Test: build wheels in test-pypi workflow

Jakob-Unfried:
This is not intended to be merged.
I just need a place to run github actions to test them.

### DIFF
--- a/.github/workflows/publish-test-pypi.yml
+++ b/.github/workflows/publish-test-pypi.yml
@@ -1,6 +1,9 @@
-name: Upload Python Package to test.pypi
+name: Build wheels and Upload Python Package to test.pypi
 
 on:
+  # on PR : TODO just for testing
+  pull_request:
+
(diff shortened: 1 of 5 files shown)


Jakob-Unfried:
Notes on what is going on here:
- The publish-test-pypi action (and I assume the same goes for the publish-pypi) did not work; the wheel was built using a wrong linux spec. This is my attempt at building proper wheels. Should also have the benefit of providing native wheels for apple silicon machines
- The action that publishes to testPyPI is supposed to run when a tag is pushed. I have set it to run on PRs temporarily, so I can test it here
- To make it work, I needed to require a newer version of setuptools, which requires python >= 3.7, so that would be a hard requirement now.
- the dependency on h5py prevents us to build wheels for some architectures (32bit windows, i686 linux). I have not tried hard to make it work for those platforms...

PR #351, opened 2024-02-29
Fix error for RAM prediction for NearestNeighbor models

p-thoma:
A `NearestNeighbor` model might not have an attribute `options`. Therefore predicting the RAM might have failed when the model was not derived from a `MPOModel` or when sites are grouped in a Simulation.


Jakob-Unfried:
Correct. Thanks!
Dont worry about the action. It just failed to upload the code coverage results, since your fork doesnt have the access rights.
I will fix the action soon, to skip that step on unmerged PRs.
Anything missing? If not I would merge?

p-thoma:
Ok perfect, thank you!

PR #354, opened 2024-03-08
Rewrite mixer_cleanup

### LINKED ISSUE #298: Rewrite Mixer Cleanup
This feature come up during the discussion in #288 

> - [ ] Rewrite `mixer_cleanup` to do the "proper" thing rather than the hack to perform another SVD sweep: do SVDs of the psi._S and absorb them into the neighboring tensors.
> A bit tricky how to update the environments, though. We could
>     1. properly update all of them, even though in most cases that's probably not needed?
>     2. just update the environments on the very left/right saved by get_resume_data() and clear the rest to be recomputed on demand.

Jakob-Unfried:
Fixes #298

There is an optimization decision regarding updating the environments.
If they are needed again, it is much cheaper to update them at this point then to recompute them later.
On the other hand, there is a descent chance they will not be needed again and updating them is unnecessary overhead.
The current implementation updates all of them.
Could introduce an argument to control the behavior?
Updating them here is O(L w chi^3).

### DIFF
--- a/tenpy/algorithms/mps_common.py
+++ b/tenpy/algorithms/mps_common.py
@@ -714,14 +714,77 @@ class Sweep(Algorithm):
         """Cleanup the effects of a mixer.
 
         A :meth:`sweep` with an enabled :class:`~tenpy.algorithms.mps_common.Mixer` leaves the MPS
-        `psi` with 2D arrays in `S`.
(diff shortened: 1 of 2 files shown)


Jakob-Unfried:
Needs better test coverage.
There are no tests that fail if I perturb the code with some choice typos.

PR #355, opened 2024-03-08
Errors on unhandled JW strings

### LINKED ISSUE #282: missing jordan-Wigner strings for some functions

Jakob-Unfried:
Fixes #282.

### DIFF
--- a/tenpy/networks/mps.py
+++ b/tenpy/networks/mps.py
@@ -4092,6 +4092,9 @@ class MPS(BaseMPSExpectationValue):
 
         i = self._to_valid_index(i)
         if isinstance(op, str):
+            if self.sites[i].op_needs_JW(op):
+                msg = 'Applying operators that need JW strings is not supported yet'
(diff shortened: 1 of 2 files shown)


Jakob-Unfried:
@p-thoma ping

jhauschild:
For reference: see also commit 74a0d4eada2ca294c3daf3c00c00e839beed5632 generalizing this a bit to actually handle JW strings by looking at charges

Issue #314, opened 2023-11-02
feature: add more consistency and error checks


jhauschild:
With Frank Pollmann and @p-thoma, we discussed today that we should add more consistency / error checks of reasonable parameters, and print big, fat warnings or even make it opt-in (by manually overriding threshold parameters) to keep running in certain instances.

Let me list a couple of examples, the list should probably be extended when implementing this:

- [x] Cylinder DMRG with way too large Ly
- [x] too large truncation errors e.g. in variational algorithms like DMRG
- [x] too large truncation errors during time evolution should stop the evolution
- [x] too large time steps for trotterization etc
- [x] Excessive use of prediction for spectral function post-processing
- [ ] spectral function: gaussian windowing too blurry? not enough t points for meaningful fourier trafo?
- [ ] there's probably more things we should check, think about what else.

Jakob-Unfried:
@p-thoma can you do a check on the time in spectral function following the pattern in #364?

p-thoma:
@Jakob-Unfried does it make sense to include a `consistency_check` call in `TimeEvolutionAlgorithm.run` (directly after 
dt is read out) ? In this way one could always warn for too large `dt` when performing a time evolution? Or move smth. like that into `Simulation.run` ?

Jakob-Unfried:
> @Jakob-Unfried does it make sense to include a `consistency_check` call in `TimeEvolutionAlgorithm.run` (directly after dt is read out) ? In this way one could always warn for too large `dt` when performing a time evolution? Or move smth. like that into `Simulation.run` ?

I decided not to, because we only need to check for small time steps if the time evolution is trotterized, and there could be evolutions which have exact evolution operators. In fact, `RandomUnitaryEvolution` is one such case

p-thoma:
> > @Jakob-Unfried does it make sense to include a `consistency_check` call in `TimeEvolutionAlgorithm.run` (directly after dt is read out) ? In this way one could always warn for too large `dt` when performing a time evolution? Or move smth. like that into `Simulation.run` ?
> 
> I decided not to, because we only need to check for small time steps if the time evolution is trotterized, and there could be evolutions which have exact evolution operators. In fact, `RandomUnitaryEvolution` is one such case.
> 
> So there currently are checks for large dt only where trotterization is actually performed, i.e. in build_U for TEBD and for MPOEvolution

Ok thanks! I see now why this is done in `calc_U` :)

jhauschild:
[...]

PR #364, opened 2024-03-13
Add consistency checks

Jakob-Unfried:
Adressing some of #314


Jakob-Unfried:
should we use a dedicated pattern for the names of those cfg:options ?

Jakob-Unfried:
This is an implementation for v0.99 that only warns. I would turn on the errors for v1.0.

PR #365, opened 2024-03-14
drop python 3.7

jhauschild:
@Jakob-Unfried is there any other file which I need to update to drop python 3.7?

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -24,7 +24,7 @@ jobs:
       max-parallel: 5
       matrix:
         # Note: python 3.12 is covered by code_coverage.yml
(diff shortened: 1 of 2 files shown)


Jakob-Unfried:
Pretty sure thats everything.
Why do we need to drop it?
I like doing it, I am just curious.

jhauschild:
No particular reason, it's just that python 3.7 has reached it's end of life, so I don't want to worry about supporting it any more, and don't think the tests need to run on python 3.7 anymore, saving resources in the github actions.

I think we can also remove our restrictions on numpy>=1.13 and scipy>=0.18, which are effectively enforced by the ABI pinning of compatible numpy anyways.

PR #373, opened 2024-03-15
Add consistency check for linear-prediction in SpectralSimulation

p-thoma:
This puts an upper bound on the relative time one can predict with linear-prediction in `SpectralSimulation` (see #314)


Jakob-Unfried:
@p-thoma 
Wanted to include this before the upcoming release, but I couldnt push to your fork.
I ended up redoing this, with fixes in the review, in #384.
Can you work with branches on the tenpy/tenpy repo in the future?

PR #390, opened 2024-03-20
Improve test_memory_usage: check numbers

Jakob-Unfried:


Jakob-Unfried:
looks like `memory_usage` is not working as expected in the github action setup.
Could be a consequence of the virtual machine for the github action runner.
It is not just a problem with pytest, as that works as expected on my local machine.

Jakob-Unfried:
Playing around with this on different machines, the reported memory usage is not stable and reproducible in a test environment, so it makes no sense to include it in the pytest suite...

PR #397, opened 2024-03-27
Skip expensive actions on "skip actions" label

Jakob-Unfried:
This is *not* intended to be merged.
These are the same changes as in #396.
This PR exist solely so I can label it with the skip actions label and check that actions actually are skipped


Jakob-Unfried:
this has fulfilled its purpose =)

PR #363, opened 2024-03-13
Charge Sector Projection

jhauschild:
this implements a function to project an initial product state *without* charge conservation into a specific charge sector in order to initialize an MPS within that sector.

PR #398, opened 2024-03-27
Continuation  Charge Sector Projection #363 

jhauschild:
This is a continuation of #363 (due to renaming of the repo/branch)

Closes #363

### DIFF
--- a/doc/changelog/latest.rst
+++ b/doc/changelog/latest.rst
@@ -18,10 +18,15 @@ Added
 
 Changed
 ^^^^^^^
-- nothing yet
+- MPS classmethod :meth:`~tenpy.networks.mps.MPS.project_onto_charge_sector` added. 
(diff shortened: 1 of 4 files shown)


jhauschild:
I think Ludwig and Philipp are already addressing this, but just as a reminder here: PurificationMPS.from_infiniteT_canonical doesn't work with two independent charges, as also pointed out in [this forum thread](https://tenpy.johannes-hauschild.de/viewtopic.php?p=1487)

p-thoma:
I just updated the method `get_charge_tree_for_given_charge_sector` which seems to have caused this bug, the way the tensors were "build" from a given charge tree was already ok for two types of charges. I also included the example given in the [Forum Thread](https://tenpy.johannes-hauschild.de/viewtopic.php?p=1487) in a test, just to be sure this works :)
@jhauschild I think this can be merged now.

PR #400, opened 2024-03-31
Fix examples

Jakob-Unfried:


p-thoma:
> Not tested yet, not clear if complete

Just updated the yaml examples that were not following the conventions for measurement functions and were therefore no longer working in v1...

Jakob-Unfried:
> Just updated the yaml examples that were not following the conventions for measurement functions and were therefore no longer working in v1...

Thanks!

PR #405, opened 2024-04-23
Qr based mpo

Jakob-Unfried:


tenpy/algorithms/mps_common.py
```diff
@@ -2392,3 +2393,89 @@ def update_local(self, _, optimize=True):
...
+        """Given a new two-site wave function `theta`, split it and save it in :attr:`psi`."""
+        i0 = self.i0
+        new_psi = self.psi
+        # old_A0 = new_psi.get_B(i0, form='A')
+        expand = self._expansion_rate(i0)
+        U, S, VH, err, renormalize = qr_theta(self.psi.get_B(i0, 'B'), self.psi.get_B(i0+1, 'B'), theta,
+                                              self.trunc_params, 
+                                              expand=expand, 
+                                              use_eig_based_svd=self.options.get('use_eig_based_svd', False),
+                                              need_A_L=False,
```
Jakob-Unfried, line 2462: You use ``U``below, so it should be ``need_A_L=True``, no?

tenpy/algorithms/mps_common.py
```diff
@@ -2392,3 +2393,89 @@ def update_local(self, _, optimize=True):
...
+    def update_new_psi(self, theta: npc.Array):
+        """Given a new two-site wave function `theta`, split it and save it in :attr:`psi`."""
+        i0 = self.i0
+        new_psi = self.psi
+        # old_A0 = new_psi.get_B(i0, form='A')
+        expand = self._expansion_rate(i0)
+        U, S, VH, err, renormalize = qr_theta(self.psi.get_B(i0, 'B'), self.psi.get_B(i0+1, 'B'), theta,
+                                              self.trunc_params, 
+                                              expand=expand, 
+                                              use_eig_based_svd=self.options.get('use_eig_based_svd', False),
```
Jakob-Unfried, line 2461: As implemented now, using ``A S B`` form, we can not support eig based svd.

I am not sure if there is a way to get around this, similar to the inversion free trick in TEBD.
If not, we should remove the ``use_eig_based_svd`` option.

If yes, we should implement that.

tenpy/algorithms/mps_common.py
MartinHefel: True, this results in an error if 'compute_err' is set to 'False'.

tenpy/algorithms/truncation.py
```diff
@@ -307,6 +307,204 @@ def svd_theta(theta, trunc_par, qtotal_LR=[None, None], inner_labels=['vR', 'vL'
...
+    renormalize : float
+    """
+    # TODO: Do we need to implement a parameter inner_labels as for the svd method?
+
+    if compute_err:
+        need_A_L = True
+
+    # Get inital guess for the right isometry
+    Y0 = _qr_theta_Y0(B_L, B_R, theta, expand, min_block_increase) # Y0: [vL, (p1.vR)]
+
```
MartinHefel, line 471: Are these labels (``B_R: [(p1.vR),vL]``) correct?

tenpy/algorithms/truncation.py
```diff
@@ -307,6 +307,204 @@ def svd_theta(theta, trunc_par, qtotal_LR=[None, None], inner_labels=['vR', 'vL'
...
+        L, U = npc.eigh(A_Ahc, sort='>')
+        S = np.sqrt(np.abs(L))  # abs to avoid `nan` due to accidentally negative values close to zero
+        U = U.ireplace_label('eig', inner_labels[0])
+    elif need_Vd:
+        U = None
+        Ahc_A = npc.tensordot(A.conj(), A, [0, 0])
+        L, V = npc.eigh(Ahc_A, sort='>')
+        S = np.sqrt(np.abs(L))  # abs to avoid `nan` due to accidentally negative values close to zero
+        Vd = V.iconj().itranspose().ireplace_label('eig*', inner_labels[1])
+    else:
```
MartinHefel, line 415: Is this else statement (if need_U=need_Vd=False) really needed? When are we only interested in the Schmidt values?

tenpy/algorithms/truncation.py
```diff
@@ -439,71 +498,148 @@ def _eig_based_svd(A, need_U: bool = True, need_Vd: bool = True, inner_labels=[N
...
+        |   -- theta --   ==>   -- t_L --- S --- t_R --
+        |      |   |                |             |
+
+    Where `t_L` is in `'A'` form and `t_R` in `'B'` form.
+
+    The decomposition for ``use_eig_based_svd=True`` is::
+
+        |   -- theta --   ==>   -- t_L --- t_R --
+        |      |   |                |       |
+
```
MartinHefel, line 518: I just realized that these schematics are wrong since the legs are combined.

tenpy/algorithms/mps_common.py
MartinHefel: ``QRBasedVariationalApplyMPO`` now supports ``use_eig_based_svd`` by using ``t_L S t_R`` tensors, where ``t_L`` and ``t_R`` are in ``'A'``, ``'B'`` or ``'Th'`` form depending on the input parameters. ``VariationalApplyMPO`` already supports the possibility to update ``new_psi`` with tensors in ``'Th'`` form.

tenpy/algorithms/truncation.py
```diff
@@ -307,6 +307,342 @@ def svd_theta(theta, trunc_par, qtotal_LR=[None, None], inner_labels=['vR', 'vL'
...
+            else:
+                t_L = npc.tensordot(A_L, U, ['vR', 'vL'])
+
+    # Compute error
+    if compute_err:
+        if use_eig_based_svd:
+            theta_approx = npc.tensordot(t_L, t_R, ['vR', 'vL'])
+        else:
+            theta_approx = npc.tensordot(t_L.scale_axis(S, axis='vR'), t_R, ['vR', 'vL'])
+        eps = npc.norm(theta - theta_approx) ** 2
```
MartinHefel, line 629: This leads to an incorrect overlap since ``theta`` is not normalized.

```suggestion
        eps = npc.norm(theta/npc.norm(theta) - theta_approx/npc.norm(theta_approx)) ** 2
```

tenpy/algorithms/mps_common.py
```diff
@@ -2392,3 +2393,101 @@ def update_local(self, _, optimize=True):
...
+        return max(expand_0 - chi / chi_max * (expand_0 - expand), expand)
+
+    def update_new_psi(self, theta: npc.Array):
+        """Given a new two-site wave function `theta`, split it and save it in :attr:`psi`."""
+        i0 = self.i0
+        new_psi = self.psi
+
+        expand = self._expansion_rate(i0)
+        use_eig_based_svd = self.options.get('use_eig_based_svd', False)        
+        t_L, S, t_R, form, err, renormalize = decompose_theta_qr_based(
```
MartinHefel, line 2459: Normalize ``theta`` prior the decomposition?

```suggestion
        theta *= 1/npc.norm(theta)
```

Jakob-Unfried:
A few notes to discuss later in person;

- [ ] Should we do a separate class that has the common code between QR-based TEBD and QR-based apply MPO?

- Signature of ``use_eig_based_svd``
  - [ ] I dont like that the forms we use depend on ``use_eig_based_svd``. It is not intuitive at all and it makes debugging harder. I think we should change the return format of ``decompose_theta_qr_based``. Maybe let it always return ``A_L, Th_L, S, Th_R, B_R``, and some of them may be None?
  - [ ] The names of the return values are not in tenpy convention: use capital single letters for tensors, e.g. ``T_L``.
  - [ ] return docstring: use same formatting as parameters. after the ``:`` should follow only the type, maybe with legs, shape info etc, but no text
  - [ ] i dont like the names ``get_left_side`` and ``need_other_side``. Mostly because the second params name does not make sense on its own. Maybe change to ``compute_t_L`` and ``compute_t_R`` ? Slightly different meanings. Depending on what we do about the return values, something again slightly different may be sensible here

- [ ] It looks like QRBasedTEBDEngine does not consider the form of the tensors returned by use_eig_based_svd. It worries me that the tests dont detect this? 

- [ ] dedicated test for decompose_theta_qr_based ? checking isometric properties, checking that the original is recovered, etc.

- [ ] In QRBasedVariationalApplyMPO. update_new_psi, I think reading out the old_B_L for decompose_theta_qr_based always in B form is not ideal. During the left sweep, we have them in A form, no?

tenpy/algorithms/truncation.py
```diff
@@ -307,133 +307,101 @@ def svd_theta(theta, trunc_par, qtotal_LR=[None, None], inner_labels=['vR', 'vL'
...
     Parameters
     ----------
-    B_L : Array with legs [vL, p, vR]
-    B_R : Array with legs [vL, p, vR]
+    T_L : Array with legs [vL, p, vR]
+        If ``move_right=True`` the tensor `T_L` must be in `'B'` or `'Th'` form.
+        If ``move_right=False`` the tensor `T_L` must be in `'A'` form.
+    T_R : Array with legs [vL, p, vR]
+        If ``move_right=True`` the tensor `T_R` must be in `'B'` form.
+        If ``move_right=False`` the tensor `T_R` must be in `'A'` or `'Th'` form.
```
MartinHefel, line 320: This is not correct yet. It might be tricky to find a convention for the leg charge, because:
- for the TEBD (where we always have `move_right=False`), psi is in `B` form (hence `theta_old=S@B_L@B_R`), and `vL_old = T_R.get_leg('vL')=B_R.get_leg('vL')` makes sense as a general convention. `vL_old` would then be the right leg of `S`.
- However, if we have `move_right=False` for the VariationalApplyMPO then `theta=A_L@Th_R`. Consequently we would have `vL_old = T_R.get_leg('vL')=Th_R.get_leg('vL')` which is the left leg of `S`.

This is a problem if `S` changes the charge.

tests/test_truncation.py
```diff
@@ -55,3 +59,95 @@ def test_truncate():
...
+    eng = mpo_evolution.ExpMPOEvolution(psi, M, options)
+    for i in range(15):
+        eng.run()
+    
+    # Fabricate dummy theta (not actually updated)
+    S = psi.get_SL(8)
+    old_T_L = psi.get_B(8, 'B').ireplace_label('p','p0')
+    old_T_R = psi.get_B(9, 'B').ireplace_label('p','p1')
+    theta = npc.tensordot(old_T_L.scale_axis(S, axis='vL'), old_T_R, ['vL','vR'])
+    theta = theta.combine_legs([['vL','p0'],['p1','vR']])
```
MartinHefel, line 101: For both cases (`move_right=False` and `move_right=True`) the tensors 'old_T_L` and `old_T_R` are in `'B'` form. It does not matter if `S` does not change the charge. However as mentioned here, this can be a problem.

tenpy/algorithms/truncation.py
```diff
@@ -534,6 +534,8 @@ def decompose_theta_qr_based(old_T_L: npc.Array, old_T_R: npc.Array, theta: npc.
 
     if compute_err:
         return_both_T = True
+    
+    theta /= npc.norm(theta) # <- TODO: discuss this, since for TEBD already normalized. Alternatively, normalize in VariationalCompression.update_local()
```
MartinHefel, line 538: We need to normalize theta somewhere if we are interested in the error. 

Does it make sense to do this in general right at the beginning, only when we need it [before calculating the error](https://github.com/tenpy/tenpy/blob/ce3ac8da60d4bf02b6a1fae5855cfd072f1c2cf9/tenpy/algorithms/truncation.py#L600) or maybe better directly in [VariationalApplyMPO.update_local()](https://github.com/tenpy/tenpy/blob/ce3ac8da60d4bf02b6a1fae5855cfd072f1c2cf9/tenpy/algorithms/mps_common.py#L2382)?

Jakob-Unfried: We take care of normalization when we decompose Xi.
There is no need to normalize theta beforehand.
Should document this.
Maybe the pictures should say something like ``theta == renormalize * T_Lc---T_Rc``.
Also, should document the ``S``  and ``renormalize`` return values accordingly.

tenpy/algorithms/truncation.py
Jakob-Unfried: - If we keep the T_L as args, should name them old_T_L, like in decompose_theta_qr_based.
- I understood from the discussion last time that we wanted to change the args of decompose_theta_qr_based, such that it no longer gets the old tensors, but rather only the info it needs, namely the old leg and the qtotal. (Should carefully document which leg we expect!) Then we can let the engine take care of getting that leg.
- If `S` changes the charge, i.e. if there is a mixer, we dont care which leg we get, the left or right leg of S.
We want to guess a leg that allows a good decomposition of the new evolved theta. We use a leg that allowed a decomposition of the old theta as a starting point and just add a few extra basis vectors (this is reasonable, since the new theta has not changed much; small time step). *Any* leg that allows a decomposition of theta will do, so it can be the left or right leg of S, it does not matter.

tenpy/algorithms/truncation.py
MartinHefel: Yes, that is true. However, the error is currently computed as `eps = npc.norm(theta-theta_approx)**2` and in general `npc.norm(theta)!=1`, but `npc.norm(theta_approx)=1`. We do get the `renormalization` from `svd_theta` or `_eig_based_svd()`, but this is not the norm of `theta`; in fact, it is a combination of `npc.norm(theta)` and `new_norm` from `truncate(S, trunc_par)`. This means to calculate the error `eps`, we still need to compute `npc.norm(theta)` somewhere. After today's meeting we concluded that it is fine to just compute the norm directly where it is needed, e.g. `eps = npc.norm(theta/npc.norm(theta)-theta_approx)**2` .

tenpy/algorithms/mps_common.py
```diff
@@ -2392,3 +2393,112 @@ def update_local(self, _, optimize=True):
...
+            It makes no sense to do this on CPU. It is currently not supported for update_imag.
+            Default is `False`.
+        compute_err : bool
+            Whether the truncation error should be computed exactly.
+            Compared to SVD-based TEBD, computing the truncation error is significantly more expensive.
+            If `True` (default), the full error is computed.
+            Otherwise, the truncation error is set to NaN.
+    """
+
+    def _expansion_rate(self, i):
```
MartinHefel, line 2437: Should we relocate this function in `truncation.py` as well, since it is just a copy of the one in [tebd.py](https://github.com/tenpy/tenpy/blob/ec96015e8801610e6fce305f102f6d44518825e6/tenpy/algorithms/tebd.py#L635)? However, it does not really fit in `truncation.py` does it?

Jakob-Unfried: no, thats fine where it is.
the clean solution would be to introduce a common superclass for all QR-based engines, but that is not worth it.

Jakob-Unfried:
@MartinHefel What is actually missing here? (feel free to edit this message)

Remaining open points
- [ ] can we compute the error in the same convention as svd_theta?
- [ ] include #437 ?

tenpy/algorithms/truncation.py
```diff
@@ -307,6 +307,330 @@ def svd_theta(theta, trunc_par, qtotal_LR=[None, None], inner_labels=['vR', 'vL'
...
+        B_R.itranspose(['vL', '(p1.vR)'])
+        Xi.itranspose(['vL', 'vR'])
+
+    # SVD of bond matrix Xi
+    if use_eig_based_svd:
+        U, S, Vd, _, renormalization = _eig_based_svd(
+            Xi, need_U=move_right, need_Vd=(not move_right), inner_labels=['vR', 'vL'], trunc_params=trunc_params
+        )
+    else:
+        U, S, Vd, _, renormalization = svd_theta(Xi, trunc_params) # <- TODO: Is it fine to not specify the charges here?
```
MartinHefel, line 583: Per default, this means `qtotal_R = Xi.qtotal` (see [here](https://github.com/tenpy/tenpy/blob/224e6cf1e7b00480e4d51047d5b0a6beb7959fa3/tenpy/linalg/np_conserved.py#L3576)). As long as `Xi.qtotal=0` it does not matter, right? Could there be a case where `Xi.qtotal!=0` and we suddenly need to care about `qtotal_LR`?

PR #410, opened 2024-05-02
v2 progress pointer

Jakob-Unfried:
The purpose the ``pointer_v2`` branch is to keep track of which changes on the main branch are already considered in the ``v2_alpha`` branch.

We ignore changes on ``main`` that affect files that do not currently exist in ``v2_alpha``.
Whenever a file is (re-)created in ``v2_alpha`` it should start from the current ``main`` version.


jhauschild:
ToDo list of things that are not merged despite the file existing in v2:

- [ ] krylov_based.py GMRES and changes there introduced by VUMPS branch. We should clean that up in the v1 branch first.

PR #413, opened 2024-05-15
Implement type checking for Config.get

Jakob-Unfried:


Jakob-Unfried:
@jhauschild do you like this approach?

Implementation of the mechanism in https://github.com/tenpy/tenpy/pull/413/commits/a360d5c4c1072413d8849f6d74e83a588f28531c

The second commit uses it wherever sensible.

PR #437, opened 2024-06-24
Qr based mpo finish up

Jakob-Unfried:


tests/test_truncation.py
```diff
@@ -123,32 +124,41 @@ def test_decompose_theta_qr_based(use_eig_based_svd):
...
+            # Check that approximated theta reproduces the original theta, up to error given by err_qr
+            if list(form) == ['A', 'B']:
+                theta_approx = renormalize_qr * npc.tensordot(T_L.scale_axis(S_qr, 'vR'), T_R, ['vR', 'vL'])
+            elif list(form) in [['Th', 'B'], ['A', 'Th']]:
+                theta_approx = renormalize_qr * npc.tensordot(T_L, T_R, ['vR', 'vL'])
+            else:
+                raise NotImplementedError
+            norm_err = npc.norm(theta_approx - theta.split_legs())
+            expect_err = np.sqrt(err_qr.eps) * npc.norm(theta)
+            assert abs(norm_err - expect_err) < 1e-6  # TODO we dont compute exactly this eps!!!
```
MartinHefel, line 144: Why not just:
```suggestion
assert err_qr.eps < tol
```
because the error is already calculated in that way ([here](https://github.com/tenpy/tenpy/blob/99124cfa7882e86ff3620c3f0bde704191179669/tenpy/algorithms/truncation.py#L609)) in theory we just have to check if `eps` really is small, right? If there is a better way to calculate the error, I would do it directly in `decompose_theta_qr_based`. Once we find "the best way" to calculate the error we can ensure that it is not accidentally changed in the future and add something like this:

```suggestion
    if list(form) == ['A', 'B']:
        theta_approx = npc.tensordot(T_L.scale_axis(S_qr, axis='vR'), T_R, ['vR', 'vL'])
    elif list(form) in [['Th', 'B'], ['A', 'Th']]:
        theta_approx = npc.tensordot(T_L, T_R, ['vR', 'vL'])
    expect_err = npc.norm(theta/npc.norm(theta) - theta_approx) ** 2
    assert abs(expect_err - err_qr.eps) < tol
```
or whatever method we settle on. Maybe I am just missing something.

Jakob-Unfried: We should not do the first option, because the point is to test if that function does what it promises, so we should not assume that it does what it promises.

We should not do the second option, in my opinion, because that is *not* the definition of truncation_error in svd_theta

PR #439, opened 2024-06-26
Purification sampling

Jakob-Unfried:
Continuation of #434

### DIFF
--- a/doc/changelog/latest/purification_sampling.txt
+++ b/doc/changelog/latest/purification_sampling.txt
@@ -1,0 +1,2 @@
+- Implement :meth:`~tenpy.networks.purification_mps.PurificationMPS.sample_measurements` for
+  :class:`~tenpy.networks.purification_mps.PurificationMPS`.
--- a/tenpy/networks/mps.py
+++ b/tenpy/networks/mps.py
@@ -3681,8 +3681,8 @@ class MPS(BaseMPSExpectationValue):
         norm_tol : float
             Tolerance
         complex_amplitude : bool
-            Do we return the amplitude (True) of the sampled bit string or the probability (False),
-            which is the |amplitude|**2.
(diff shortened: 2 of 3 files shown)


Jakob-Unfried:
One more thing i just realized: the purification version returns probabilities, the MPS version complex weights.
That is not ideal...

Ideas how to unify this? @jhauschild @sajantanand ?

sajantanand:
I couldn't think of a way to do this since when sampling the purification MPS, it seemed most natural to accumulate the probability from the RDMs on each site (whether they are for the physical leg or the ancilla leg). While one could calculate a complex weight in the case of sampling both physical and ancilla leg, I don't know if this would be useful.

Then again, if one samples the ancilla leg, the probability returned isn't useful either.

Jakob-Unfried:
Reminder to merge
https://github.com/tenpy/tenpy_notebooks/pull/4
and commit the submodule update

Jakob-Unfried:
I agree that it makes sense for the purification version to work with probabilities. i am just wondering how to avoid confusion. there is a string expectation that the return value has the same meaning as in the superclass.

maybe we could add an optional kwarg that controls what should be returned, complex amplitude or probability? if the default is "complex amplitude", we retain backwards compatibility for the existing MPS functionality. the purification version would just raise in tjat case and tell you to put "probability"

@jhauschild ?

sajantanand:
I added the functionality you described.

Jakob-Unfried:
Great! If the actions complete, this is ready to merge =)

Jakob-Unfried:
Turns out the notebook takes too long to run to be useful in pytest.
I marked most cells to be skipped by pytest.
The one non-trivial cell that does run guarantees that 
a) the sampling code runs without error
b) the distance between the observables is not too large

For future reference, any notebook cell with the tag `'skip-execution'` is skipped by ``pytest --nbmake``.

PR #444, opened 2024-07-16
Log and store trunaction errors for `RealTimeEvolution`

p-thoma:
Currently the truncation error is computed, but neither logged or saved for a `RealTimeEvolution`.
This PR logs the `TruncationError` (accumulated at least for eps and latest for ov) in the `run_algorithm` method of the `RealTimeEvolution` class - the logging could also be done at the level of the `TimeEvolutionAlgorithm` but logging the error for one individual step might be more useful here? 
In the `RealTimeEvolution` class, the ov and eps errors are also recorded as measurements (added to the `default_measurements` of that class).
Furthermore, the redundant (and confusing) logging line before running the engine in the `run_algorithm` method was removed, the same information is already logged by the `TimeEvolutionAlgorithm`.

Lastly, as some measurements can take quite a while, I thought it might be useful to also report the time it took for all measurements to complete (this was done in `make_measurements` of the `Simulation` class).

### DIFF
--- a/tenpy/simulations/simulation.py
+++ b/tenpy/simulations/simulation.py
@@ -659,8 +659,11 @@ class Simulation:
     def make_measurements(self):
         """Perform measurements and merge the results into ``self.results['measurements']``."""
         self.logger.info("make measurements")
+        m_start_time = time.time()
         results = self.perform_measurements()
         self._merge_measurement_results(results)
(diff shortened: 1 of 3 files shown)


jhauschild (approved):
Looks good, thanks!

PR #451, opened 2024-07-30
Adjust v1 examples

Jakob-Unfried:
Still making more changes to the examples that generate the figures for the v1 publication...


Jakob-Unfried:
merge #485 instead

PR #464, opened 2024-09-12
minor fix of consitency_check to allow imaginary TDVP time evolution

jhauschild:
@sajantanand suggested this. 
Can you please check that it works as I think it should for what you want to do? If you give your okay, I can merge it into main.


sajantanand:
Apparently, lambda arguments cannot have parenthesis (I get an error `SyntaxError: Lambda expression parameters cannot be parenthesized`). After I remove the parenthesis in the arguement, everything works.

However, it would be nice to have the evolved imaginary time show up in the logging. I have changed the logger in `run` from `algorithms/algorithm.py` to print out both the real and imaginary parts of the evolved time. However, this will affect the logging of all real-time evolution too. Do you have a better suggestion for how to do this?

PR #467, opened 2024-10-10
Handle zero blocks correctly in truncated `qr`

attila-i-szabo:
I've noticed that `npc.qr` with a cutoff breaks if any of the blocks disappear completely. This PR fixes this behaviour:
* `tools.math.qr_li` returns early if no diagonal elements are kept, before a next QR decomposition would lead to LAPACK errors
* `qr` skips such empty blocks from `q_data`, `r_data` so they don't lead to errors down the line
* and filters the legcharges so they only refer to the nonzero blocks

As far as I can tell, this breaks no currently functioning code. Let me know what tests etc. should be added.


Jakob-Unfried:
See #475

PR #471, opened 2024-11-13
Add Molecular Hamiltonian

bartandrews:
I would like to contribute the Molecular Hamiltonian to the selection of TeNPy models. This is a commonly used model in quantum chemistry applications, for which TeNPy has the useful property of natively supporting N and Sz conservation.

This PR is related to qiskit-community/ffsim#334

### DIFF
--- a/doc/changelog/_latest.rst
+++ b/doc/changelog/_latest.rst
@@ -7,3 +7,7 @@ The contents are auto-generated from multiple files.
 
     Contents are modified by ``stitch_changelog_latest`` in ``doc/conf.py``
     Any ``.txt`` file in ``doc/changelog/latest/`` is included verbatim.
+
+
+- Add the :class:`~tenpy.models.molecular.MolecularModel` class for quantum chemistry applications.
+
--- a/doc/changelog/_latest.rst
+++ b/doc/changelog/_latest.rst
@@ -7,7 +7,3 @@ The contents are auto-generated from multiple files.
 
     Contents are modified by ``stitch_changelog_latest`` in ``doc/conf.py``
     Any ``.txt`` file in ``doc/changelog/latest/`` is included verbatim.
-
-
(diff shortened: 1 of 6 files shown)


Jakob-Unfried:
Looks good to me!

Please remove the changes to `_latest.rst`.
Without those, I am happy to merge.

One optional nag:
Maybe it would be good to check in the test if the couplings are added correctly.
The logic to add them is complicated and typo-prone.
E.g. by creating the Hamiltonian on O(3) sites once via the model and once via brute-force numpy and comparing?
This is just to make sure that the model actually does what the docstring says it does.

Just FYI: the linter didnt like your type hints because `Config` was not imported.
Since we dont have them consistently yet, it doesnt really matter if they are included here.

bartandrews:
Thank you Jakob for the review! I have removed the changes to `_latest.rst` (although it still shows up as a change on github for some reason...maybe there is some whitespace). I agree that adding the couplings in the MolecularModel is complicated and typo-prone. The main reason for this is because we were trying to be smart about the for loops, using e.g. `itertools.combinations` and `plus_hc` whenever possible to reduce the size of the loops. Without these tricks/optimizations, the for loop logic is actually quite simple. Therefore, I have added a simple MolecularModel in the test file (without these for loop tricks) and I check that the MPOs produced are the same. I have also added back the type hints, thank you for spotting that

bartandrews:
@Jakob-Unfried Do you know when the next release of TeNPy is due, so that we can use this MolecularModel in Qiskit?

Jakob-Unfried:
@jhauschild

we can do a minor release essentially whenever. ill try to close some more of the almost finished PRs and then push a release,
say by the end of the week.

note though that we have some persistent problem with the conda forge setup and new releases dont make it to conda forge at the moment, only to PyPI.

PR #473, opened 2024-12-03
fix expected output in the c_mps_mpo example (6 sites, not 4)

Dobrzyniecki:
The example describes a N=6 lattice, and the output of the code is `<Sz> = [ 0.5 -0.5  0.5 -0.5  0.5 -0.5]`, not `<Sz> = [ 0.5 -0.5  0.5 -0.5]`.


Jakob-Unfried:
You are right, thanks.

However, in your fix you put a dot `.` where a space should be though.
I cant simply amend this, since i cant push to your branch.
If you want the honors of authoring the commit yourself, please correct that, otherwise i can push the same fix directly to the main repo.

Dobrzyniecki:
ouch, good catch, thank you.

Jakob-Unfried:
Thanks!

PR #475, opened 2024-12-06
Handle zero blocks correctly in truncated `qr`

Jakob-Unfried:


Jakob-Unfried:
Looks good to me on reading this, hut I would prefer it with a test that fails on current implementation but passes here.

PR #477, opened 2024-12-10
Bug in MPO.to_TermList(...) fixed

MaSC5023:
If the variable `max_range` is equal to $M$, then the longest coupling term is $OP_0*OP_M$ (taking `i`$=0$ as an example). Therefore, the range of `k` should be from $0$ to $M$.
For instance, the original function produces no output for the `tf_ising` model, while my modification resolves this issue.

### DIFF
--- a/tenpy/networks/mpo.py
+++ b/tenpy/networks/mpo.py
@@ -1049,7 +1049,7 @@ class MPO:
                 continue
             partial_L[self.get_IdL(i)] = [([], 1.)]
             if self.finite:
-                max_range = min(max_range, L - i)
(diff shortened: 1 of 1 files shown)


Jakob-Unfried:
Thanks!

Looks correct to me on reading it.
I would prefer to have a simple test that fails before the test and passes now.
Sounds like you have some code snippets that do this?
Do you want to push a test yourself? Otherwise, could you just paste these code snippets here?

Jakob-Unfried:
Looks good. I added a test in #479 that currently fails. Your changes should fix it =)
Thanks!

PR #478, opened 2024-12-10
Change recommended citation to v1 codebase paper

Jakob-Unfried:


Jakob-Unfried:
@jhauschild Do you agree?
Should we also put it in docs/literature?

PR #479, opened 2024-12-16
Fix MPO to term list (one off error in max_range)

Jakob-Unfried:
Adding a test to #477.


MaSC5023:
> Adding a test to #477.

Thanks for adding the test! Looks great to me—let's merge it!

PR #481, opened 2025-02-14
Fix consistency checks with illegal values

Jakob-Unfried:


Jakob-Unfried:
all done, assuming checks pass

PR #483, opened 2025-02-18
Improve accessibility of models for newcomers

Jakob-Unfried:


Jakob-Unfried:
Does not render as intended yet...

Jakob-Unfried:
- [ ] point to `add_local_term`

jhauschild (approved):
As said in the comment, the XXZ model can always conserve Sz, hence I added an extra field hx to demonstrate how one selects charge conservation depending on parameters.

Issue #176, opened 2022-08-31
Generalize ExponentiallyDecayingCouplingTerms


jhauschild:
Right now, the ExponentiallyDecayingCouplingTerms add couplings with a uniform decay rate lambda.
It should be fairly straight-forward to allow `lambda_` to be a non-uniform array with one value per site of the MPS (unit cell).
This could be used for e.g. a SSH-type model with long range, see https://tenpy.johannes-hauschild.de/viewtopic.php?t=302

~~Slightly annoying, right now the convention is that the strength is ~ lambda^|i-j|, which includes a lambda on the first site.
We should probably change this to something like  prod(i < k < j) lambda[k] , which amounts to lambda^|i-j-1| for uniform couplings.
This might break backwards compatibility with existing code using the ExponentiallyDecayingCouplingTerms. I assume that this would not affect many users, though.~~

PR #491, opened 2025-03-17
Non uniform decay rates

coxipi:
This implements the non-uniform decay rates discussed in #176. `lambda_` can now be a 1D array with the size of the MPS `L`.  I've added a very basic test for `bc = "finite"`.

Looking at other closed PRs, it seems maybe I should have not included the changelog .txt description? Let me know, I can remove it.

If a scalar is given, I could have transformed `lambda_` to a 1D array early on for a simplified implementation, but I'm assuming that if a scalar is given, it should remain as such in the class attribute `exp_decaying_terms`.


Jakob-Unfried:
We just discussed this and will fix the follwing issues:
@jhauschild 

- fix to_TermList
- adjust test: we expect ``[l[0], l[0] * l[2], l[0] * l[2] * l[4], l[2], l[2] * l[4], l[4]]``
- doc meaning of lambda[i]

PR #513, opened 2025-09-03
Investigate behavior of TEBD error

Jakob-Unfried:
We got a report of weird behavior via email.
Essentially, the truncation error does not seem to improve with increasing bond dimension, for the QRBased TEBD

### DIFF
--- a/tenpy/algorithms/tebd.py
+++ b/tenpy/algorithms/tebd.py
@@ -681,15 +681,27 @@ class QRBasedTEBDEngine(TEBDEngine):
         old_B_L = self.psi.get_B(i0, 'B')
         old_B_R = self.psi.get_B(i1, 'B')
 
+        compute_err = self.options.get('compute_err', True, bool)
         _, S, B_R, form, trunc_err, renormalize = decompose_theta_qr_based(
             old_qtotal_L=old_B_L.qtotal, old_qtotal_R=old_B_R.qtotal, old_bond_leg=old_B_R.get_leg('vL'),
             theta=theta, move_right=False,
             expand=expand, min_block_increase=self.options.get('cbe_min_block_increase', 1, int),
             use_eig_based_svd=self.options.get('use_eig_based_svd', False, bool),
             trunc_params=self.trunc_params,
-            compute_err=self.options.get('compute_err', True, bool),
+            compute_err=compute_err,
             return_both_T=False,
         )
+        if compute_err:
+            chi_max = self.trunc_params.get('chi_max', None)
+            chi_current = len(S)
+            if trunc_err.eps > 1e-16 and chi_max is not None and chi_current < chi_max:
+                msg = ('QRBased decomposition resulted in large truncation error even though the bond '
+                    'dimension is not maxed out yet. Try increasing the expansion rate, e.g. '
+                    'via the `cbe_expand_0` and `cbe_min_block_increase` options for the engine. '
+                    'You probably should compare to a "regular" (non QR-based) engine and see how '
+                    'fast the bond dimension needs to grow for your scenario. '
+                    'See https://github.com/tenpy/tenpy/pull/513 .')
+                warnings.warn(msg, stacklevel=2)
         assert form[1] == 'B'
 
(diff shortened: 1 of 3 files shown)


Jakob-Unfried:
I have added a test that runs TEBD for the model / setting of interest.
See plots below.
Indeed it looks like something odd is going on.

First thought is that the bond dimension for QR-TEBD does not grow fast enough, so maybe it is enough to just increase ``cbe_expand``.
Will investigate further...


Screenshot 2025-09-03 at 17 13 27


Screenshot 2025-09-03 at 17 14 46

Jakob-Unfried:
Indeed, if i push the expansion rate to allow rapidly growing bond dimension, we get what looks like well-behaved expected behavior to me:

This is with
```python
tebd_params['cbe_expand'] = .5
tebd_params['cbe_expand_0'] = .5
tebd_params['cbe_min_block_increase'] = 5
```

Screenshot 2025-09-03 at 18 05 17

Jakob-Unfried:
Btw this was the test code that I since removed

```python
@pytest.mark.parametrize('use_large_expansion_rate', [True, False])
def test_QRBasedTEBD_warning_on_too_low_expansion(use_large_expansion_rate,
                                                  plot=False,
                                                  chis=[50, 100, 200],
                                                  L=8, N=20, Nmax=20, U=40, dt=.7, N_steps=100):
    """See :issue:`513`.

    If entanglement grows rapidly during time evolution, the default expansion rate in QRBasedTEBD
    might not be sufficient to let the bond dimension grow fast enough.
    After a user ran into this problem and could not figure out why, we added a warning to make
    it clear whats going on. This test checks if the warning is issued.
    """
    model_params = dict(L=L, t=1, U=U, mu=0, n_max=Nmax, bc_MPS='finite', filling=N/2, conserve='N')
    model = BoseHubbardChain(model_params)

    if use_large_expansion_rate:
        # there should be no warnings
        context = nullcontext()
    else:
        # the large time step causes rapid entanglement growth, too fast for the default
        # expansion rate to keep up.
        # check if we get the expected warning in this scenario.
        context = pytest.warns(UserWarning, match='QRBased decomposition resulted in large truncation error')

    ov_errs_per_chi = {}
    eps_errs_per_chi = {}

    actual_chi_per_chi = {}
    for chi in chis:
        psi = MPS.from_lat_product_state(model.lat, [[N], [N], [0], [0]])
        trunc_params = dict(chi_max=chi, svd_min=1e-14, trunc_cut=None)
        tebd_params = dict(dt=dt, order=1, N_steps=1, trunc_params=trunc_params)
        if use_large_expansion_rate:
            tebd_params['cbe_expand'] = .5
            tebd_params['cbe_min_block_increase'] = 5
        eng = tebd.QRBasedTEBDEngine(psi, model, tebd_params)
        errs = []
        actual_chis = []

        with context:
            for n in range(N_steps):
                eng.run()
                errs.append(eng.trunc_err)
                actual_chis.append(max(psi.chi))

        ov_errs_per_chi[chi] = np.array([err.ov_err for err in errs])
        eps_errs_per_chi[chi] = np.array([err.eps for err in errs])
        actual_chi_per_chi[chi] = np.array(actual_chis)

    if plot:
        # this was useful while looking into whats going on, but should not be part of
        # automatic testing
        import matplotlib.pyplot as plt
        fig, (ov_ax, eps_ax, chi_ax) = plt.subplots(3, sharex=True)
        for chi in chis:
            ov_ax.semilogy(ov_errs_per_chi[chi], label=f'chi_max={chi}')
            eps_ax.semilogy(eps_errs_per_chi[chi], label=f'chi_max={chi}')
            chi_ax.plot(actual_chi_per_chi[chi], label=f'chi_max={chi}')
        fig.suptitle('Larger expansion rate' if use_large_expansion_rate else 'Default expansion rate')
        ov_ax.set_ylabel('ov_err')
        eps_ax.set_ylabel('eps')
        chi_ax.set_ylabel('chi')
        chi_ax.set_xlabel('Trotter steps')
        ov_ax.legend()
        eps_ax.legend()
        plt.show()
```

PR #520, opened 2025-10-03
Single site bug

Jakob-Unfried:
Continuation of #508

> This PR address two bugs in DMRG, found by Stefan Divic.
> 
> 1. If using `chi_list` to perform sequential DMRG sweeps with increasing bond dimension, no `chi_max` is set for the environment sweeps. Thus, these default to `chi_max=100` and severely truncate the MPS. Now, if one is doing `chi_max` and `optimize=False` (env sweeps), I set `chi_max` to be the maximum bond dimension of the current chi.
> 2. In SingleSiteDMRG, `mixer_activate` failed if `mixer = None`. I added to the if statement to avoid this.


Jakob-Unfried:
LGTM. Thanks @sajantanand

PR #527, opened 2025-10-06
Magnetic unit cell commensuration

Jakob-Unfried:
Fixes the problem addressed by #452.

As it turns out, the symmetric gauge was simply not correctly implemented.
Additionally, in all cases where the implementation is correct, the model params ``mx, my`` that claim to control the
magnetic unit cell do not have any effect at all, since they did not change the hopping amplitudes actually realized in the coupling model; The periodicity imposed by the gauge choice means that *after tiling* during ``add_coupling`` the same prefactors end up being used.

As noted by @pelegemanuel in the original PR, the definition of the symmetric gauge only gets a phase of ``pi`` if we shift by ``q`` (the denominator of the flux per plaquette), and thus the size of the magnetic unit cell is actually ``(2 * q, 2 * q)``. This means that we need systems with periodic boundaries to be commensurate with that cell.

However, even if the system is commensurate, the current implementation of the symmetric gauge leads to tiling the phases from the generated ``(q, q)`` patch, which essentially means that plaquettes that go across a periodic boundary had an erroneous extra pi flux. It thus accidentally implements a different model.
This is what @pelegemanuel detected in their ED study.

This PR:
1. Adds a modification of @pelegemanuel's ED test to verify the Hofstadter models via their spectra from small system ED. This test fails for the symmetric gauge if any bc is periodic, highlighting the bug.
This test should also convince us that the following changes dont introduce new errors in the model
3. Refactors the code that generates the phases
4. Adds extra tests that make sure the phases give rise to the correct flux on all plaquettes

### DIFF
--- a/doc/changelog/latest/pr_527.txt
+++ b/doc/changelog/latest/pr_527.txt
@@ -1,0 +1,4 @@
+- Fixed a bug in the :class:`~tenpy.models.hofstadter` models, where the symmetric gauge was not
+  implemented correctly and yielded a different physical model. Removed the config options for
+  controlling the magnetic unit cell size, since in all well-configured cases they had no effect.
+  See :pull`527` for details.
--- a/tenpy/models/hofstadter.py
+++ b/tenpy/models/hofstadter.py
@@ -12,113 +12,151 @@ from .lattice import Square
 from ..networks.site import BosonSite, FermionSite
 from .model import CouplingMPOModel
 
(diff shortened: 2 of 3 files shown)


Jakob-Unfried:
I am pretty sure the proposed condition is too strict. if there are multiple flux quanta per MUC, the hoppings might repeat with a smaller period

PR #564, opened 2025-10-28
Fix qtotal during mixing

### LINKED ISSUE #424: Make sure S has no qtotal during mixing

Jakob-Unfried:
Fixes #424

### DIFF
--- a/tenpy/algorithms/dmrg.py
+++ b/tenpy/algorithms/dmrg.py
@@ -906,8 +906,11 @@ class TwoSiteDMRGEngine(DMRGEngine):
             )
             S_a = S
         else:
-            qtotal_LR = [self.psi.get_B(i0, form=None).qtotal,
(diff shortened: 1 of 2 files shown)


Jakob-Unfried:
It seems like mixer (not sure if both?) are currently incompatible with `diag_method='ED_all'`.
This is because we assume that the qtotal does not change when setting the `qtotal_LR` to be the same as before
at

https://github.com/tenpy/tenpy/blob/2712de56aef8791c16717033dbfcb53b9d3939a7/tenpy/algorithms/dmrg.py#L909-L914

but with `diag_method='ED_all'` the qtotal of theta changes (in general), so this choice of `qtotal_LR` can lead to a charged S.

maybe we should instead do ``qtotal_LR = [qtotal_old_B_L, qtotal_theta - qtotal_old_B_L]`` ?

PR #594, opened 2026-01-05
Fix two site tdvp explicit phc

Jakob-Unfried:
Follow up to #593


Jakob-Unfried:
I am now looking into if this should also apply to the zero-site updates

PR #609, opened 2026-01-28
Normalize MPS tensors during canonical_form_finite

### LINKED ISSUE #596: Stability issue in canonical_form for small norm states (e.g., Gutzwiller projection)
**The Problem**

I have encountered a numerical stability issue when restoring the canonical form (with a finite `cutoff`) of an MPS that has a very small physical norm (e.g., < 1e-15).
This situation typically arises after applying a Gutzwiller projection.
When `canonical_form()` is subsequently called, the internal SVD steps encounter tiny singular values.
As a result, standard absolute cutoffs often discard physically relevant singular values, or the routine fails.

**Possible solutions**

We (TeMFpy team, i.e. @attila-i-szabo and me) have two ideas on how one could handle this: 

- Add a flag that allows for normalization during QR decomposition (at each step to handle to handle underflow). For this one could add a version of the QR decomposition like this (maybe with an inplace option):

```python
def qr_normalized(M, **kwargs):
    """Normalises the matrix M and returns its former norm and QR decomposition."""
    norm = npc.norm(M)
    Q, R = npc.qr(M / norm, **kwargs)
    return Q, R, norm
```

- Add an option to truncate according to relative weight (this may be a good feature i.g.)

We are happy to implement the changes, but we need your opinion on how and where they should be implemented.

Jakob-Unfried:
Fixes #596
Continuation of #608

### DIFF
--- a/tests/test_mps.py
+++ b/tests/test_mps.py
@@ -1012,12 +1012,14 @@ def test_fixes_596(renormalize):
     psi.form = [None] * L
     psi.test_sanity()
 
-    psi.canonical_form_finite(cutoff=1e-10, renormalize=renormalize)
+    cutoff = 1.0e-10
+    psi.canonical_form_finite(cutoff=cutoff, renormalize=renormalize)
     psi.test_sanity()
     expect_norm = 1 if renormalize else 0.5**L
-    assert abs(psi.norm - expect_norm) < 1e-16
-    assert abs(psi.overlap(psi) - expect_norm**2) < 1e-16
-    assert psi.overlap(psi_old) > (1 - 1e-8) * expect_norm
+    assert abs((psi.norm - expect_norm) / expect_norm) < 10 * cutoff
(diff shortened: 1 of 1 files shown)


tests/test_mps.py
```diff
@@ -1000,5 +1000,25 @@ def test_fixes_600_hdf5(tmp_path):
...
+
+    for tensor in psi._B:
+        tensor /= 2
+    psi.form = [None] * L
+    psi.test_sanity()
+
+    psi.canonical_form_finite(cutoff=1e-10, renormalize=renormalize)
+    psi.test_sanity()
+    expect_norm = 1 if renormalize else 0.5**L
+    assert abs(psi.norm - expect_norm) < 1e-16
```
jhauschild: This just checks that psi.norm  < 1.e-16. You should have a relative value comparison `abs((psi.norm - expect_norm)/expect_norm)`  here (and the line below), not absolute, to ensure the correct prefactor 2^-L of the norm.

PR #622, opened 2026-03-04
Measure and reset MPS

sajantanand:
Add function similar to `sample_measurements` that measures (a subset of) sites of an MPS and resets them to the resulting product state. This is needed for trajectory simulations of quantum channels.


Jakob-Unfried:
- I find the comments on infinite MPS and the comparison to sample_measurement a bit confusing
- One test I can think of is to compare with a numpy representation, where we can do the projection explicitly, e.g. explicitly writing down the projector as a matrix and contracting, of course that only works for finite bc
- For infinite I think we could do some examples where we can work out the results on paper, simplest case would be GHZ state, maybe with different unit cell sizes

PR #625, opened 2026-05-21
Add ArnoldiEvolution as alternative to LanczosEvolution in TDVP

jonasreimann:
In order to enable Lindblad evolution using quantum trajectories for systems with long-ranged interactions, one needs TDVP to work for non-Hermitian Hamiltonians. The current bottleneck prohibiting this is the `LanczosEvolution` inside TDVP which assumes hermitian matrices. See also [this forum post](https://forum.tenpy.org/viewtopic.php?t=401) . Here I try to resolve this by introducing the following changes:

- With the help of an LLM I have implemented the class `ArnoldiEvolution` on top of the existing `Arnoldi` in exact analogue to `LanczosEvolution`. A newly introduced parameter `hermitian` which is stored in `lanczos_params` toggles the use of `ArnoldiEvolution` instead of `LanczosEvolution` if `hermitian=False`.

- Another ingredient for the quantum trajectories algorithm is tracking the norm of the state. Currently this was not done in TDVP, as far as I could tell. I have modified both `TwoSiteTDVPEngine` and `SingleSiteTDVPEngine` to correctly track the norm in `psi.norm`, in accordance with the flags `preserve_norm` and `normalize`.

I added corresponding tests for `ArnoldiEvolution` to the testsuite. Additionally, the attached example script tests the time evolution of a minimal non-hermitian BoseHubbard model using both TDVP and TEBD. See attached plots. They show the convergence behavior w.r.t $\chi,\Delta t$, the density distribution over time and the norm of the state, comparing TDVP against TEBD.    
As the differences between TEBD and TDVP have similar order of magnitude as the numerical error stemming from the convergence parameters, I concluded that they agree well.

bh_convergence
bh_density
bh_tebd_vs_tdvp
[non_hermitian_bose_hubbard_comparison.py](https://github.com/user-attachments/files/28110631/non_hermitian_bose_hubbard_comparison.py)

### DIFF
--- a/tenpy/algorithms/tdvp.py
+++ b/tenpy/algorithms/tdvp.py
@@ -275,7 +275,12 @@ class TwoSiteTDVPEngine(TDVPEngine):
         else:
             theta = theta.combine_legs([['vL', 'p0'], ['p1', 'vR']], new_axes=[0, 1], qconj=[+1, -1])
         qtotal_i0 = self.psi.get_B(i0, form=None).qtotal
-        U, S, VH, err, renorm = svd_theta(theta, self.trunc_params, qtotal_LR=[qtotal_i0, None], inner_labels=['vR', 'vL'])
+        U, S, VH, err, renorm = svd_theta(
+            theta,
+            self.trunc_params,
+            qtotal_LR=[qtotal_i0, None],
+            inner_labels=['vR', 'vL'],
+        )
         self.psi.norm *= renorm
         B0 = U.split_legs(['(vL.p0)']).replace_label('p0', 'p')
         B1 = VH.split_legs(['(p1.vR)']).replace_label('p1', 'p')
--- a/tenpy/algorithms/tdvp.py
+++ b/tenpy/algorithms/tdvp.py
@@ -36,7 +36,7 @@ import logging
 import warnings
 
 from ..linalg import np_conserved as npc
-from ..linalg.krylov_based import LanczosEvolution
+from ..linalg.krylov_based import ArnoldiEvolution, LanczosEvolution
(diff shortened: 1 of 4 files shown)


Jakob-Unfried:
Looks reasonable. Thanks for the contribution!
I have not gone into detail yet.

1. Make sure the actions pass.
2. We should have tests for a run of non-hermitian TDVP. I am reasonably convinced by your comparison that the method works in principle, but we should include it as part of the automatic tests, also to make sure we will not break it in the future. 
   - We should have it analogously to the existing case, comparing to TEBD. If possible, this should be with a less computationally demanding system than your example
   - To really trust it, i would love to see a comparison to ED
   - Is there a simple toymodel you can use, with known observable dynamics, that we can check?

jonasreimann:
Thanks for the feedback!
I have added a non-hermitian config to the existing TDVP test which should cover non-hermitian TDVP+ArnoldiEvolution in the testsuite. 

Also, I have put together a simple comparison to ED for the Hatano-Nelson model with asymmetric hoppings. See attached script and the resulting plot. They also confirm the correct behavior of non-hermitian TDVP.

Unfortunately, it turned out difficult to find a test case with low computational effort which shows TDVP + LanczosEvolution failing, but TDVP + ArnoldiEvolution succeeding. It seems to require large bond dimensions and large effective Hamiltonians to make the difference between both methods visible. Let me know if you also need one such example, then I could dig deeper.
test_hatano_nelson
[test_hatano_nelson.py](https://github.com/user-attachments/files/28477180/test_hatano_nelson.py)

Jakob-Unfried:
LGTM now.
I will give @jhauschild a few days to veto, otherwise I am happy to merge.

PR #627, opened 2026-06-02
Fixed from_drop_charge string argument bug

### LINKED ISSUE #626: BUG from_drop_charge(...) fails for String Arguments
When passing a string as the charge argument onto the class method `LegCharge.from_drop_charge`, the interpreter raises a ValueError stating that the charge is not contained within `chinfo.names`. That is because it tries to access the charge entry after it has already been removed in `chinfo = ChargeInfo.drop(leg.chinfo, charge)`: 

```@classmethod
    def from_drop_charge(cls, leg, charge=None, chargeinfo=None):
        r""" [...] """
        if charge is None:
            return cls.from_trivial(leg.ind_len, chargeinfo, leg.qconj)
        chinfo = ChargeInfo.drop(leg.chinfo, charge)    # <--- CHINFO DOESN'T CONTAIN CHARGE
        if chargeinfo is not None:
            assert chinfo == chargeinfo
            chinfo = chargeinfo
        if isinstance(charge, str):
            charge = chinfo.names.index(charge)    # <--- RAISES ERROR, BECAUSE CHINFO DOES NOT HAVE CHARGE
        return cls.from_qind(chinfo, leg.slices, np.delete(leg.charges, charge, 1), leg.qconj)
```
The presumable reason this bug went unnoticed is because it is only called once in `Array.drop_charge` by an integer argument, for which the bug-triggering if clause is not called. A minimal fix is to change the line `charge = chinfo.names.index(charge)` to `charge = leg.chinfo.names.index(charge)`. 

On an unrelated matter, I also want to suggest adding the methods `LegCharge.from_drop_dipolar_charge` and `Array.drop_dipolar_charge` for completeness.

KarolexDev:
Closes #626

### DIFF
--- a/tenpy/linalg/charges.py
+++ b/tenpy/linalg/charges.py
@@ -914,12 +914,14 @@ class LegCharge:
         """
         if charge is None:
             return cls.from_trivial(leg.ind_len, chargeinfo, leg.qconj)
+        if isinstance(charge, str):
+            charge = leg.chinfo.names.index(charge)
+        if isinstance(leg.chinfo, DipolarChargeInfo) and charge not in leg.chinfo._dipole_idcs:
+            raise KeyError ("In the case of dipolar charges all of the dipolar components of must be dropped.")
         chinfo = ChargeInfo.drop(leg.chinfo, charge)
         if chargeinfo is not None:
             assert chinfo == chargeinfo
             chinfo = chargeinfo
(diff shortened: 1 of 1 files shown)


Jakob-Unfried:
Thanks!

Looks good, we can merge if the actions pass.

Issue #558, opened 2025-10-21
BUG: Adiabatic Continuity while charge pumping in fractional Chern insulator models.


andrewkhardy:
**Describe the bug**
Hello, thanks to the TenPy community for supplying so many examples. I believe however I have encountered
an issue (not sure if it necessarily a bug) in charge pumping for fractional states.  I am attaching the example code from [Prof. Pollman's Boulder summer school lectures](https://drive.google.com/drive/folders/1iiSjyLpPdc4__FoK39wwJNqtlBH6xvfl?usp=share_link) for the IQH Haldane model with the output when I run locally. 

**Minimal example or code extract**

```python
import matplotlib.pyplot as plt
import numpy as np
from tenpy.algorithms import dmrg
from tenpy.networks.mps import MPS
from tenpy.models.haldane import FermionicHaldaneModel

def plot_results(data):
    plt.figure()
    ax = plt.gca()
    ax.plot(data['phi_ext'], data['QL'], marker='o')
    ax.set_xlabel(r"$\Phi_y / 2 \pi$")
    ax.set_ylabel(r"$ \langle Q^L(\Phi_y) \rangle$")

    plt.figure()
    ax = plt.gca()
    colors = ['b', 'g', 'r', 'c', 'm', 'y', 'k']
    color_by_charge = {}
    for phi_ext, spectrum in zip(data['phi_ext'], data['ent_spectrum']):
        for q, s in spectrum:
            q = q[0]
            label = ""
            if q not in color_by_charge:
                label = "{q:d}".format(q=q)
                color_by_charge[q] = colors[len(color_by_charge) % len(colors)]
            color = color_by_charge[q]
            ax.plot(phi_ext * np.ones(s.shape),
                    s,
                    linestyle='',
                    marker='_',
                    color=color,
                    label=label)
    ax.set_xlabel(r"$\Phi_y / 2 \pi$")
    ax.set_ylabel(r"$ \epsilon_\alpha $")
    ax.set_ylim(0., 8.)
    ax.legend(loc='upper right')

def run(model_params, dmrg_params, prod_state, phi_ext=np.linspace(0, 1.0, 21)):
    data = dict(phi_ext=phi_ext, QL=[], ent_spectrum=[])
    print("Initial state: ", prod_state)
    eng = None
    for phi in phi_ext:
        print("=" * 100)
        print("phi_ext = ", phi)

        model_params['phi_ext'] = phi

        if eng is None:
[...]

PR #633, opened 2026-07-14
Fix t1 sign in Haldane FCI example

iarayad:
The sign of t1 in the bosonic Haldane FCI example means that the flat Chern band is the higher energy band, not the lowest one. Changing the sign inverts the bands, which is desired to observe the FCI at 1/4 total filling. This PR fixes that typo.
This issue was reported in a thread of #558 for example.


Jakob-Unfried:
Thanks!

Jakob-Unfried:
I will rebase first to get a fresh run of the actions with recent bugfixes on main included

PR #637, opened 2026-07-28
Release V1.1.1

Jakob-Unfried:


Jakob-Unfried:
[Last attempt to run the TestPyPI action](https://github.com/tenpy/tenpy/actions/runs/30372998253) came up with some errors when testing the built wheels

- linux wheels all pass
- windows wheels resulted in worrisome errors: mismatching legs in qr based decomposition
- mac wheels: test_vumps[2-SingleSiteVUMPSEngine-False] fails with too large of an error. I assume the difference is because of different blas implementation or something of the sort.

Will investigate more later

PR #640, opened 2026-08-18
Fix docs

Jakob-Unfried:


Jakob-Unfried:
These warnings actually point to unwanted behavior, links inserted where they are not supposed to, e.g.:

Screenshot 2026-08-30 at 17 43 08

Screenshot 2026-08-30 at 17 44 40

Screenshot 2026-08-30 at 17 46 23

PR #630, opened 2026-07-12
Fixed Bug in tenpy.networks.site.GroupedSite

IanROsborne:
Fixed a bug in GroupedSite which would throw a 
```shell
TypeError: 'bool' object is not iterable
```
error when sites with different charge_to_JW_parity values were grouped. 

Example: 

```
ferm_site = SpinHalfFermionSite(cons_N='N', cons_Sz= 'Sz' )
spin_site = SpinHalfSite(conserve= 'Sz')
set_common_charges([ferm_site, spin_site], new_charges= 'same')
sites = [ferm_site, spin_site]

site = GroupedSite(sites, labels = ['e', 'i'] , charges = 'same')
```

Returns:
```shell
TypeError: 'bool' object is not iterable
``` 

Also moved the location ```c2JWps``` closer to where it is used for ease of read.

PR #646, opened 2026-08-30
Fix bug in GroupedSite

Jakob-Unfried:
Continuation of #630


Jakob-Unfried:
Note: Failing Sphinx build is a known issue, addressed in #640 (and currently also failing on main)

